PR #37, opened 2021-02-10
Full audio

clapierre:
Feel free guys to request improvements as I did my best to change audiobook to full audio where I could.  There was a reference to an audiobook in ONIX 81-01 but since that says audiobook I left it as it.

### DIFF
--- a/UX-Guide-Metadata/principles/index.html
+++ b/UX-Guide-Metadata/principles/index.html
@@ -95,7 +95,7 @@
 				retrieves all books with metadata of <code>&lt;link rel="dcterms:conformsTo"
 					href="http://www.idpf.org/epub/a11y/accessibility-20170105.html#wcag-a" /&gt;</code> or the
 				equivalent for wcag-aa and ultimately for wcag-aaa would fill this need.</p>
-			<p>Audiobooks created for mainstream use provide important access for many users with disabilities even
+			<p>Full audio books created for mainstream use provide important access for many users with disabilities even
 				though they are not accessible to others. Book providers may already have search tools aimed at finding
 				audiobooks, since they are popular and fairly common, but if not, this would be a good addition to the
 				suite of tools for finding accessible books. Note that audiobooks may not pass WCAG requirements because
@@ -149,7 +149,7 @@
 			<h2>Order of Key Information</h2>
 			<ol type="1">
 				<li><p><a href="#screen-reader-friendly">Screen Reader Friendly</a></p></li>
-				<li><p><a href="#audiobook">Audiobook</a> (if present)</p></li>
+				<li><p><a href="#full-audio">Full Audio</a> (if present)</p></li>
 				<li><p><a href="#accessibility-summary">Accessibility Summary</a></p></li>
 				<li><p><a href="#accessibility-conformance">Accessibility Conformance</a></p></li>
 				<li><p><a href="#certified-by">Certified By</a></p></li>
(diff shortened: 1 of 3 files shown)


GeorgeKerscher:
I wonder if it would help to clarify that we mean the entire content of the title is in an audio recording, and hence full audio. This may go below the place where it says: NOTE
As long as you find <meta property="schema:accessModeSufficient">auditory</meta> the book can claim “Full Audio: Yes”.

UX-Guide-Metadata/principles/index.html
```diff
@@ -95,7 +95,7 @@ <h2>Discovering Accessible Content</h2>
 				retrieves all books with metadata of <code>&lt;link rel="dcterms:conformsTo"
 					href="http://www.idpf.org/epub/a11y/accessibility-20170105.html#wcag-a" /&gt;</code> or the
 				equivalent for wcag-aa and ultimately for wcag-aaa would fill this need.</p>
-			<p>Audiobooks created for mainstream use provide important access for many users with disabilities even
+			<p>Full audio books created for mainstream use provide important access for many users with disabilities even
```
mattgarrish, line 98: Just a phrasing nitpick, but even with the extra space it still sort of sounds like a full "audiobook".

Could this maybe be something like: "Books with full audio" ?

It would need changing in a number of other spots if acceptable.

UX-Guide-Metadata/techniques/epub-metadata.html
```diff
@@ -155,29 +155,29 @@ <h5>UI</h5>
...
 				<p class="note">As long as you find <code>&lt;meta
 						property="schema:accessModeSufficient"&gt;auditory&lt;/meta&gt;</code> the book can claim
-							<q><b>Audiobook: Yes</b></q>.</p>
+							<q><b>Full Audio: Yes</b></q>.</p>
 				<p>There may also be other <code>accessModeSufficient</code> entries with combinations of access modes
 					such as <q>textual, visual</q> which indicate other ways to read the book.</p>
-				<p class="note">If you can not report that this is an audiobook then do not present any information to
-					the user for this category, since 99% of EPUB books are not audiobooks. Simply omit this
+				<p class="note">If you can not report that this is a full audio book then do not present any information to
+					the user for this category, since 99% of EPUB books are not considered full audio ebooks. Simply omit this
```
mattgarrish, line 180: I'd strip ", since 99% of EPUB books are not considered full audio ebooks". It's not really adding anything and avoids getting into percentages.

UX-Guide-Metadata/techniques/onix.html
```diff
@@ -55,10 +55,10 @@ <h2>ONIX Metadata</h2>
 			</p>
 			
 			<p class="note">
-				ONIX messages describe products for the global book supply chain and will be sent from publisher or creator of the epub or digital audiobook to those who will make the products available for sale, lending or subscription.
+				ONIX messages describe products for the global book supply chain and will be sent from publisher or creator of the epub or digital full audio books to those who will make the products available for sale, lending or subscription.
```
mattgarrish, line 58: Isn't this instance actually referring to audiobooks since it talks of them separately from epub?

UX-Guide-Metadata/techniques/onix.html
```diff
@@ -171,9 +171,9 @@ <h4>ONIX record describing an EPUB</h4>
 				</section>
 				<section>
 					<h4
-						id="onix-record-describing-an-audiobook-here-is-an-example-of-an-onix-record-version-3.0-for-describing-an-audiobook."
-						>ONIX record describing an Audiobook Here is an example of an ONIX record (version 3.0) for
-						describing an audiobook.</h4>
+						id="onix-record-describing-an-full-audio-book-here-is-an-example-of-an-onix-record-version-3.0-for-describing-a-full-audio-book"
```
mattgarrish, line 174: A couple of "an full" instances in this paragraph that should be "a full"

mattgarrish (changes requested):

UX-Guide-Metadata/principles/index.html
clapierre: Changes made and pushed.

UX-Guide-Metadata/techniques/epub-metadata.html
clapierre: Removed this statement.

UX-Guide-Metadata/techniques/onix.html
clapierre: @gregoriopellegrino left this for you.

UX-Guide-Metadata/techniques/onix.html
clapierre: Fixed.

clapierre:
@GeorgeKerscher you asked

> I wonder if it would help to clarify that we mean the entire content of the title is in an audio recording, and hence full audio. This may go below the place where it says: NOTE
As long as you find auditory the book can claim “Full Audio: Yes”.

I agree with you George except in this section we are only looking at the metadata inside the OPF file so how will they know its a entire content of the title is an audio recording, that. what triggers the accessmodesufficient = "auditory" which is where we are here telling the consumer of the metadata if they see "auditory" alone they can claim "Full Audio: Yes" without having to parse the entire ebook to figure that out.  What you are recommending is for how the producer of the ebook can determine if the book should contain this metadata or not.  I think its important but maybe not for this document.

laudrain:
Today I heard from someone at the EU Commission using the expression "spoken book" to explain full audiobook.
I find this interesting and wanted to bring it to your appreciation.
IMO it helps distinguishing audio in ebooks between full text recorded (spoken book) and other audio media.

gregoriopellegrino:
Ok in the ONIX techniques I've fixed the example for the full-audio and reintroduced the audiobook example.

I've a doubt on the audiobook technique for ONIX, with respect to the other techniques (since we don't have a reference in the principles document):
- which value should be displayed? Now I have `Yes / (if No - Omit this section)`
- what the UI should display? Now I have `Audiobook: Yes`

avneeshsingh:
@gregoriopellegrino if ONIX code indicate that it is an audio book, then one should show 
Audiobook: Yes, as you mentioned. And if ONIX codes are not able to confirm audio book or a text sync audio book then nothing is displayed i.e. Audiobooks: label is also not displayed.

gregoriopellegrino:
With my last commit it is complete on my side. I think we can merge.

gregoriopellegrino (approved):

Issue #43, opened 2021-04-05
accessibilityHazard:none incorrect UI


clapierre:
In Section 2.8.5 when the accessibilityHazard metadata is = "none" the UI should not say "None provided" because there was metadata provided.

It should say 
Omit this Section
or
Hazards: none
or
Hazards: No Flashing, No Sound, No Motion Simulation.

gregoriopellegrino:
I would prefer `Hazards: none`

GeorgeKerscher:
I am OK with none as well.

avneeshsingh:
I think we can make this change in user experience guide.

clapierre:
The only issue here is that the reason we added "Omit this section" as an option was for the following.

Nothing presented to the user because there are no hazards; since 99.9% of EPUBs won’t have any hazards, there is no need to report Hazards: None for every EPUB.

Do we still feel this way or we really want to see Hazards: none for every EPUB?

avneeshsingh:
It looks that the principles document is in-sync with our above discussions it says:
Unlike other accessibility properties, the presence of hazards can be expressed either positively or negatively. This is because users are more often looking for content that is safe for them. This section should always be displayed; indicate that no metadata was provided if that is the case.
https://w3c.github.io/publ-a11y/UX-Guide-Metadata/principles/#hazards

But in techniques we suggest to omit hazards if it is none. I am leaning towards following principles document.

avneeshsingh:
To summarize:

If there is a hazard like flashing hazard then UI should convey it:
Hazard: Flashing

If metadata reports No Flashing, No Sound, No Motion Simulation then UI can convey:
Hazards: No Flashing, No Sound, No Motion Simulation.
or
Hazards: none

If metadata is not provided or hazards is explicitly mentioned as "unknown in metadata". Then UI should convey:
Hazards: unknown
Another way can be to convey "unknown" if hazard metadata has value "unknown" and convey "none provided" if hazard metadata is missing. But I think that this detail will not make much difference for the users. So, in favor of simplicity, we can stick to displaying "unknown" in both the cases.

PR #46, opened 2021-06-23
Update on Hazards to address #43

clapierre:
EPUB metadata examples updated for Hazards none, unknown as discussed.


GeorgeKerscher:
I think I did this correctly. I feteched the pull request and then reviewed the techniques file. I looked at the section on hazards. I see that the use of none was perfectly clear. Also, it was clear if no hazard metadata was present. I hope I got all this correce.

gregoriopellegrino (approved):

PR #48, opened 2021-07-19
Abstract proposal

gregoriopellegrino:
I wrote this abstract taking some texts of mine and of Charles, do you think it could be good?

It is a starting point, we can modify everything.

### DIFF
--- a/UX-Guide-Metadata/principles/index.html
+++ b/UX-Guide-Metadata/principles/index.html
@@ -42,7 +42,11 @@
 	  </script>
 	</head>
 	<body>
-		<section id="abstract"></section>
+		<section id="abstract">
+			<p>Finding, buying, and finally reading a book is a very personal experience. It can be challenging for users with disabilities to understand whether a certain publication meets their reading needs (before buying, borrowing or downloading it).</p>
+			<p>The accessibility characteristics of publications must be found easily and there must be a consistent way to display this technical metadata in a user-friendly and consistent way.</p>
+			<p>This document proposes a shared framework for presenting publication accessibility metadata in a user-friendly manner.</p>
+		</section>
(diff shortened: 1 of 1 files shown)


clapierre (approved):
:) looks familiar ;)

GeorgeKerscher:
I think the abstract looks fine. I envision other marketing materials will provide modivation to readers. This is short and to the point. We do not want to go into more debth here.
I approve.

clapierre (changes requested):
From email thread, suggested improvements.

Finding, buying, and finally reading a book is a very personal experience.  For most of us this is routine where we go to a bookstore search for the title we are interested in or perhaps browse the best sellers list, then purchase and start reading the book.

Now consider you are blind and rely on assistive technology to read this book.  Will my screen reader work with this title? Are there image descriptions that will be spoken to describe these images, are there page numbers which are accessible, is the reading order correct so I don't hear chapter 4 before chapter 3?  These are just a few of the accessibility concerns consumers have when trying to purchase and ultimately read electronic publication.

The good news is more and more publishers are creating Born Accessible publications and getting their workflows certified as accessible by trusted organizations.

This document proposes a shared framework for presenting publication accessibility metadata in a user-friendly manner.

avneeshsingh:
"The good news is more and more publishers are creating Born Accessible publications and getting their workflows certified as accessible by trusted organizations."

It looks very specific. Better to say:
The good news is more and more publishers are creating Born Accessible publications and also getting the accessibility validation or audit done by independent organizations.

gregoriopellegrino (commented):
updated! please review it

clapierre (approved):
Looks good, all changed incorporated.

PR #49, opened 2021-07-22
removed unneeded image potential copyright issue

clapierre:


gregoriopellegrino (commented):
Do we need to change the images used? Or do we proceed without any problems?

Perhaps we could use the Noun Project (and then cite it in the sources): https://thenounproject.com/

mattgarrish:
If they're public domain images, you're okay, otherwise you have to follow any attribution requirements and restate the licensing.

Even if you got permission to use them, you may want to check what W3C licensing stipulates. It may allow use the owner didn't intend if you don't say otherwise.

PR #50, opened 2021-07-22
Matts suggested changes

clapierre:
These were changes suggested by Matt on 7/16.

### DIFF
--- a/UX-Guide-Metadata/principles/index.html
+++ b/UX-Guide-Metadata/principles/index.html
@@ -88,10 +88,7 @@
 			<p>While every user has different accessibility needs, meeting the Web Content Accessibility Guidelines
 				(WCAG) 2.0 requirements, even at level A, is a baseline that ensures a book will be widely accessible.
 				Book providers may therefore wish to create specific search capabilities to permit users to find books
-				that have declared conformance to WCAG at any level. A search filter called <q>Accessible</q> that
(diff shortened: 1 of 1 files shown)


gregoriopellegrino (commented):
The changes for the text for me are very good.

I have a doubt about the links to the implementations: it seems to me that we decided not to put them, so that in the future we could add techniques, without having to modify the principles... But maybe I'm wrong...

avneeshsingh:
@gregoriopellegrino  and @mattgarrish 

In EPUB accessibility 1.0 under IDPF we made similar decision.
But current version of EPUB accessibility 1.1 has pointers to specific text of techniques.

It would be good to have a uniform approach for all the accessibility documents. 
If we want to avoid pointers to specific text of techniques in user experience guide, then it has more compelling reason for EPUB accessibility 1.1, because the abstract layer of EPUB accessibility will become a recommendation, and we can update the recommendation by AC vote, so it will be pretty stable document up to next version. But EPUB accessibility techniques is a WG note which can be updated by the WG without AC vote.
EPUB accessibility 1.1
https://w3c.github.io/epub-specs/epub33/a11y/index.html

mattgarrish (approved):

gregoriopellegrino (approved):

avneeshsingh:
Good to see approval from both Matt and Gregorio, both are in-sync.
Ok from me also.

PR #57, opened 2021-09-09
Added references to ONIX record near each example

gregoriopellegrino:
As we mentioned in the call, for each XPath example (in the ONIX techniques) I've included the reference to the ONIX record, so that you can quickly understand where the data is extracted from.


clapierre (approved):
Looks great!

avneeshsingh (approved):

PR #59, opened 2021-09-16
Additional proofread

mattgarrish:
Mostly minor rewording suggestions in the earlier talkative sections of the guide.

The document jumped back and forth from "publication" to "book", so I standardized the language on "publication" since that's typically what we do to avoid book bias.

The only other slightly substantive change I've made is to move up the list of links in the section that describes the ordering. I find it helps to place it between the explanation of why there is ordering and the snippets that introduce the fields (e.g., so if you're in a hurry you can skim the list and move on).

But have a review and let me know if you disagree with any of it.


clapierre (approved):
Looks great Matt as usual.

gregoriopellegrino (approved):

PR #90, opened 2022-04-14
Broke up the massive table into separate tables.

clapierre:
Also cleaned up scoping and whitespace of the table code


editadapt:
Many thanks!

GeorgeKerscher:
I reviewed the document and the break-up of the tables looks good. I think it is good to merge.

PR #103, opened 2022-05-23
Added two list items about character limits

GeorgeKerscher:
In the Some High-level consideration section, I added two list items about character limits.


clapierre (commented):
I think this statement needs to be reworked "Our guideline is between 1,000 and 2,000 characters."  As this implies a 500 character summary wouldn't be acceptable.

Maybe something like "Ideally the accessibilitySummary should be under 1000 characters and would recommend against an accessibility summary over 2000 characters in length."

PR #109, opened 2022-05-26
Fixes to language bullet

mattgarrish:
Sorry, I didn't have a chance to look at the recently merged pull request until just now. Opening this new PR to offer a few suggested edits:

- I think it would be helpful to separate the recommendation for a single summary to a separate bullet and put it first.
- The text said "language of the primary publication" but it sounds like this should be "primary language of the publication".
- I've added bullet about declaring metadata language since there's no harm in mentioning it even if any publication conforming to the accessibility standard will have to set it.

I triggered some reformatting of tag indenting and lines in the commit so in the interest of easier review this bullet:

> - The language of the Accessibility Summary should be in the same language of the primary publication. For example, a publication in French should have the accessibility Summary also in French. While some publications may have many passages in another language, the primary language should be determined. For example, a publication intended to teach Spanish to English speaking students would have the primary language as English. There should only be one accessibility summary; do not produce the accessibility summary in multiple languages.

is now:

> - There should only be one accessibility summary. Do not include multiple accessibility summaries in different languages.
> - The language of the accessibility summary should be in the primary language of publication. For example, a publication in French should have the accessibility summary also in French. While some publications may have many passages in another language, the primary language should be determined. For example, a publication intended to teach Spanish to English speaking students would have the primary language as English.
> - Ensure the language of the accessibility summary is declared in the metadata.

### DIFF
--- a/drafts/schema-a11y-summary/index.html
+++ b/drafts/schema-a11y-summary/index.html
@@ -61,9 +61,11 @@
 				or this information can be retrieved from ONIX <a href="https://ns.editeur.org/onix/en/196">code list:
 					196</a>; Code: 00: Accessibility Summary.</p>
 
-			<p class="note">Refer to the <a
-					href="https://www.w3.org/2021/09/UX-Guide-metadata-1.0/principles/">User Experience Guide for Displaying Accessibility Metadata 1.0</a> for libraries and retailers.</p>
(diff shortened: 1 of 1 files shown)


clapierre (approved):
I like the improvements Matt!

PR #113, opened 2022-05-29
Updated the Screen Readerly Friendly section

GeorgeKerscher:
I added a paragraph after the heading to explain Screen Reader Friendly. I removed the two rows and have only one here now.


gregoriopellegrino (approved):

clapierre (commented):
"Schema" should be changed to "Schema.org" as this is the standard way of referencing this piece of metadata.

clapierre (approved):

PR #116, opened 2022-06-02
Added two list items about no markup and jargon

GeorgeKerscher:
This commit addes two list items in section 3. The no markup is one and the no jargon is the other. The no jargon is similar to the use plain language list item, but I think it works.


clapierre (commented):
> It should be noted that markup in metadata is commonly not allowed. While it may be tempting to use headings, bullets, bold, or  other markup, this should not be used.

I think since it is disallowed we should make this strong as "this must not be used." instead of "this should not be used."

> The use of jargon or language that is not commonly understood should be avoided. It is better to explain  what is meant . For example, the term "reflow" may not be understood by most people. 

Extra spaces between "explain  what" and "meant ."

PR #118, opened 2022-06-06
Proposal for pageSource property

mattgarrish:
The problems with our use of dc:source to identify the source of pagination has come up again in discussions around the accessibility metadata. The dc:source element is too generic, as we discovered between 3.0 and 3.0.1, and the hack of a "source-of" property is as doomed as the `refines` attribute. It also makes for confusion when there is no source, as you can't say that a source of something is nothing and that nothing is the source of the pagination. We're hacking on hacks.

This PR is a proposal to create an `a11y:pageSource` property to finally deal with this issue properly. It removes the ambiguity of dc:source and makes more sense when used with a value like "none" for digital-only editions.

The preview is available here: https://raw.githack.com/w3c/publ-a11y/feature/page-source-id/drafts/page-source-id/index.html

### DIFF
--- a/drafts/page-source-id/css/common.css
+++ b/drafts/page-source-id/css/common.css
@@ -1,0 +1,45 @@
+
+/****************************************************/
+/*  property tables                                 */
+/****************************************************/
+
+/* ensure confomity of width for property tables */
+table.tabledef {
+	border-spacing: 0px;
+	border: none;
+	font-size: 1em;
+	width: 100%
+}
+
+table.tabledef td, table.tabledef th {
+	border: none;
+	background-color: rgb(236,246,255);
+	color: rgb(0,0,0);
+	padding: 0.3em;
+	vertical-align: top;
+}
+
+table.tabledef th {
+    text-align: left;
+	vertical-align: top;
+    width: 8em;
+    padding-left: 1em;
+}
+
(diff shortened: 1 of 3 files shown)


clapierre (commented):
Matt, instead of the property being "pageSource" what do you think about "paginationSource" or "pageBreakSource"?

GeorgeKerscher:
I love this specification. I found one misspelling. Where it says:
In this example, the pagination correponds to a print edition with an ISBN number. A URN is 
misspelled: corresponds

clapierre:
I think if we are adding a new "pageBreakMarkers", I think we should be consistent and have "pageBreakSource" or more explicitly "pageBreakMarkerSource" so it is super clear these two items are related.

clapierre:
Sounds good I am ok with "pageBreakSource"

GeorgeKerscher:
I am good with this as we

GeorgeKerscher:
This looks good.

GeorgeKerscher (approved):

mattgarrish:
After the discussion on the last call, I've removed the a11y prefix from the property. It's replacing the dc:source/source-of properties, so it doesn't need to be in the a11y namespace, which to date we've only used for accessibility reporting metadata.

PR #122, opened 2022-07-14
Remove the one line remember the summary  may be the only metadata th…

GeorgeKerscher:
I removed one line that says this may be the only metadata the user sees. This undermines the implementation of the presentation of other accessibility metadata. …e user finds


clapierre (approved):
I guess this makes sense given how things are moving.

PR #127, opened 2022-07-28
Add note about potential ambiguity in naming of tts reading options

mattgarrish:
This pull request adds a note to the TTS section to address the issue of a "read now" TTS option sounding similar to "read aloud books" full audio publication. The naming doesn't change the fundamental concepts of a full audio publication versus TTS synthesis.

This PR addresses the final issue that was raised when we resolved to publish the document, so merging will also gives us the green light to publish.

### DIFF
--- a/drafts/audio-playback/index.html
+++ b/drafts/audio-playback/index.html
@@ -122,9 +122,9 @@
 				full audio are never as simple to load and play as pure audiobooks are because there are control files
 				that a simple audio playback device does not understand.</p>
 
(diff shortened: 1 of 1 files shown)


mattgarrish:
> add at the end of this sentence as compared with pre-recorded human speech.

I almost wrote something like that, but it doesn't have to be pre-recorded human speech. You could do synchronize higher-quality TTS or even mix both (body narrated by human and back matter by TTS). I figured the on-the-fly TTSing was enough without confusing people with these kinds of nuances.

PR #128, opened 2022-08-31
deleted parts we agreed on and added an abstract

GeorgeKerscher:
I hope this can be merged so we can review it on the A11Y sub-task force of the Accessibility Summary.


clapierre (approved):
looks like a good start George.

PR #129, opened 2022-09-01
Sept 1 changes added to this version

GeorgeKerscher:
I made the changes discussed on the September 1 call. I think I got it right.


clapierre (approved):
looks like all the suggested changes from 9/1/22 meetng

PR #134, opened 2022-10-19
Formatting tables

gregoriopellegrino:
I've formatted the tables with CSS and removed "accessibilityFeature" on each row of the second table. It seems much more intelligible to me now.

### DIFF
--- a/drafts/schema-a11y-summary/index.html
+++ b/drafts/schema-a11y-summary/index.html
@@ -38,6 +38,43 @@
 				}
 			};
 		</script>
+		<style type="text/css">
+			/* Table zebra style... */
+
+			table.zebra {
+				font-size:inherit;
+				font:90%;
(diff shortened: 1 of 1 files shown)


clapierre (commented):
Everything looks great Gregorio except the feature is just "index" not "linked index" right?
all others listed are the real accessibilityFeature values so I think we must keep it consistent.  Now we can specify linked index maybe in another column.  not sure how to handle that.

GeorgeKerscher:
An index being present is a feature, and if it has links to the top of the page or to a specific spot on the page is worth adding to the summary. Yes, be consistent that all names would have exact correliations to other features.

gregoriopellegrino:
I've changed _Linked index_ to _index_

clapierre (approved):
Looks good!

PR #142, opened 2023-01-06
Modified the accessibilitySummary definition and rationale in the dra…

GeorgeKerscher:
In the repo, there is a drafts folder. I modified the principles where the old definition and rationale were located. I made those two changes.…ft folder


clapierre (approved):

gregoriopellegrino (commented):
Fine for me!

If merge, then I proceed with the other two moves that we defined in the editors' call.

PR #166, opened 2023-06-19
Ux guide: Book Features and Charts, Diagrams and Formulas

gregoriopellegrino:
Here is my proposal for the Charts, Diagrams and Formulas and Book Features section (which right now is more of a placeholder, I think we should find a better solution for this metadata, maybe some we can move to Additional Accessibility Information).

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -234,6 +234,26 @@
                 </ul>
 			</section>
 			
+			<section id="graphics-and-formulas">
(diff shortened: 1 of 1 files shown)


clapierre:
Since you are using class="example" for the aside should we also include role="doc-example" for compleatness?

gregoriopellegrino:
> Since you are using class="example" for the aside should we also include role="doc-example" for compleatness?

Pinging @mattgarrish since I don't know if it works with ReSpec.

Also I'm not sure if `doc-example` can be applied to an aside.

PR #170, opened 2023-07-05
Improving Hazards and merging Book features in Additional Metadata

gregoriopellegrino:


gregoriopellegrino:
Improved with:
- added a long example for hazards
- added the accessibility summary section
- added a bullet point to additional accessibility information for conformance and certification

# Hazards examples
In case there are known hazards, long version:
```
The publication contains content that rapidly changing lights or visual stimuli, as well as simulated movements, which can cause discomfort, distraction, photosensitive seizures, or motion sickness.
```

# Accessibility Summary
The Accessibility Summary is a of the accessibility that complements, but does not duplicate, the other discoverability metadata.

It is a free-form field that allows authors to describe the additional information to the accessible properties of the resource.

Due to its nature, no specific processing of the content is required; it is sufficient to extract the text from the metadata and display it to end users.

Example:
```
Subject experts were used to create the ALT text. The tables are represented as images, and the full text of the table is provided below the image within the details element, which can be expanded. There are several short videos in Amerigan Sign Language that explains key concepts.
```

PR #173, opened 2023-07-07
Refine Non visual & adds Pre recorded audio

gautierchomel:


gregoriopellegrino (approved):

clapierre (approved):
Just wondering instead of "can be consulted in text" what about "can be consumed in text"
but I can live with "consulted" just seems a bit awkward to me.

PR #180, opened 2023-07-11
Editorial cleanup

mattgarrish:
Let me know if any of you have any issues with what I've done. I've tried to standardize all the sections as best I could so they read similarly.

(Probably the most changed section is the conformance once. I added explanations to all the "simple statements", for example, to clarify how they apply.)

I committed some changes to main before starting to edit to cleanup the markup, line lengths, etc., so not every change will show in the file changes. It's probably better to read the [preview](https://raw.githack.com/w3c/publ-a11y/principles/editorial-pass/UX-Guide-Metadata/draft/principles/index.html).

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -107,7 +107,7 @@
 		</section>
 		<section id="sotd"></section>
 		<section id="general-overview">
-			<h2>General Overview</h2>
+			<h2>General overview</h2>
 
 			<p>This document helps those who wish to render accessibility metadata directly to users understand how to
 				represent machine-readable accessibility metadata in a user-friendly User Interface / User Experience
@@ -162,7 +162,7 @@
 			</section>
 		</section>
 		<section id="discovering-accessible-content">
-			<h2>Discovering Accessible Content</h2>
+			<h2>Discovering accessible content</h2>
 
 			<p>The guidelines for presenting accessibility metadata detailed in this document are intended to improve
 				the user experience when readers browse the catalog entry for a publication. However, accessibility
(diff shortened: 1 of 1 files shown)


mattgarrish:
Noting that there were no examples for the additional information section so I put in a placeholder. Might help to have at least one.

avneeshsingh:
Looks good!
Some improvements are:
The examples of supports non-visual reading say:
Readable in audio and braille.
Not fully readable in audio and braille.
May not be fully readable in audio and braille.

We should give an indication that we are talking about synthetic audio here so that people are not confused with pre-recorded audio.

Second piece is of links to techniques of pre-recorded audio. It states the following, so it should be corrected:
EPUB Accessibility: Enable Reading Without Sight
ONIX: Enable Reading Without Sight


If we have no other comments up to 15 UTC today, then you can merge it to draft principles. I will send the link to draft principles in the agenda.

gregoriopellegrino (approved):
Great work, thank you!

mattgarrish:
I'll fix the links to point to the full audio section. That appears to be the most relevant section, although it can always be changed later.

clapierre (commented):
I like the changes Matt.
My only concern is that if an EPUB has been third party certified who have a credential which is a Logo that information is buried under the additional accessibility information and not easily shown on the primary page.
I would hate to have to say if you want to find out if a book is GCA certified go to the Accessibility Page then click on "Additional accessibility Information"
We should have an example where if a publication is 3rd party certified like LIA or GCA this information can be displayed on the primary page.

mattgarrish:
@clapierre you should probably open issues for the changes you want. I can't really make modifications like those in a cleanup pass.

PR #186, opened 2023-10-11
Addedcertifier information and credentials after statement

GeorgeKerscher:


gregoriopellegrino:
The changes are fine with me. In LIA's proposal we suggested using the terms: "This publication is certified as compliant with accessibility (minimum, medium, maximum) standards"

But I think it's about those small changes that can occur at the implementation level.

PR #187, opened 2023-10-20
Proposal for general overview

gregoriopellegrino:
As agreed I worked on the first part of the document, on one hand to align it with the new structure, another to give some information that we know, for example:
- order of key accessibility information
- possibility of flexibility implementers
- accessibility summary info

This is a proposal for which I expect comments, in view of our call on Monday.


UX-Guide-Metadata/draft/principles/index.html
GeorgeKerscher, line 1: I reviewed and it looks good. I expect that when we reviewed the merged material that we will need to wordsmith some of the language. However, overall it looks good and ready to merge as far as I can tell.

gautierchomel: Fine to me.

gautierchomel (approved):

clapierre (commented):
Looks good, merged

PR #188, opened 2023-10-23
 Made the changes discussed today

GeorgeKerscher:
I made the changes we discussed today. I also made some minor changes where I saw language issues and these were very small. One problem I see is that the conformance section does not follow the same pattern as the rest of the sections. Everywhere else there is a descriptive statement and then the compact. All the conformance statements are compact and the detailed conformance information is where there are different approaches. I don't know if we want to make changes here or not. The conformance section is a bit different. Anyway, take a look and let us know what you think.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -128,7 +128,7 @@
 
 			<p>The accessibility summary was intended (in EPUB Accessibility 1.0) to describe in human-readable
 				prose the accessibility features present in the publication as well as any shortcomings. 
-				Starting in  EPUB Accessibility version 1.1 the accessibility summary became a human-readable summary of the accessibility that complements, but does not duplicate, the other discoverability metadata.</p>
+				Starting in  EPUB Accessibility 1.1 the accessibility summary became a human-readable summary of the accessibility that complements, but does not duplicate, the other discoverability metadata.</p>
 
 			<p>This document offers guidance on how to aggregate and display metadata to end users; these are not strict guidelines, but suggestions for providing a consistent experience for users through different portals. Different implementers may choose to implement these guidelines a slightly different way, some examples can be seen in the <a href="#implementations">Implementations</a> section of the document.</p>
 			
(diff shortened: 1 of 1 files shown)


clapierre (approved):
looks good George

avneeshsingh:
Looks good. I will merge it tomorrow, after 8 UTC, if I did not see any important comments for further improvements.

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -190,19 +190,20 @@ <h4>Examples</h4>
...
+								electronic braille.</li>
 						</ul>
 					</aside>
 
 					<aside class="example" title="Compact explanations">
 						<ul>
-							<li>Readable in computer-generated audio and braille.</li>
-							<li>Not fully readable in computer-generated audio and braille.</li>
-							<li>May not be fully readable in computer-generated audio and braille.</li>
+							<li>Readable in computer-generated audio and refreshable braille.</li>
```
mattgarrish, line 203: This example uses "refreshable braille" but the next two use "electronic braille". I know the note says they're interchangeable, but it would be good to be consistent within any one set of example text.

Also, should these examples say "readable as" or "readable using"? Something about "readable in" makes it sound like the publication is providing the audio or braille, but maybe it's just me.

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -571,16 +571,16 @@ <h4>Examples</h4>
...
 
-					<p>The "<a href="#additional-accessibility-information">Additional accessibility information</a>"
-						section is first shown as descriptive statements followed by an example that shows a single
+					<p>The detailed conformance information section is first shown as descriptive statements followed by an example that shows a single
 						compact statement that covers the whole conformance group. Either the descriptive or compact
 						style is appropriate, but do not provide both approaches.</p>
 
-					<aside class="example" title="Additional Accessibility Information with descriptive statements">
-						<p>This publication meest accepted accessibility standards.</p>
+					<aside class="example" title=" descriptive statements with detailed conformance information">
```
mattgarrish, line 578: "descriptive" in the title attribute should be capitalized.

mattgarrish, line 578: It's also unclear what "descriptive statements" refers to here. From the old title, it sounds like it was describing the text of the Additional Accessibility Information section. But now it seems to be describing the two paragraphs before it. But when I read the next example, it says it has compact statement yet the paragraphs before the section are exactly the same. I think we're trying to make some distinction about the way the conformance information is written, but having only the diffs to review I'm not sure what it is. I think the titles need clarification, though.

PR #192, opened 2023-10-29
Updated the conformance section

GeorgeKerscher:
I made the changes as we discussed on the call. I also replaced the word requirement with standard in several locations.  Whenever we refer to conforming to a standard I included EPUB Accessibility and WCAG.


gautierchomel:
I mishandled something and it resulted with merging the PR. With my apologizes.

PR #205, opened 2023-11-02
Update User Experience Guidelines title to 1.1

rickj:
Update User Experience Guidelines title to 1.1

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -2,7 +2,7 @@
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en" xml:lang="en">
 	<head>
 		<meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
Hi, I think a cross editor discussion is needed. In my opinion this revision of the guidelines is not backward-compatible with the previous one (the logic behind it has changed) so for me it would be an update of a major version (e.g. 2.x).

GeorgeKerscher:
I agree that this is a major revision. The Accessibility Summary alone is enough for the change, but also the language used to describe the various entries has also changed.

rickj:
The intent of my pull request was to get it off of 1.0 as it is causing confusion with internal and publisher contacts when I try to discuss the work that is going on

gregoriopellegrino:
Closing, because we're updating to versione 2.0

PR #208, opened 2023-11-13
Small improvements

gregoriopellegrino:
As discussed in the last call I have:

- simplified the sentence about the accessibility summary in the general overview
- fixed the note about what to show in case of absence of metadata
- for each key information I inserted a note to indicate what to do in case of absence of metadata
- added an example for _Supports nonvisual reading_ in case of absence of metadata


gregoriopellegrino:
Merging to have a clearer overview

PR #210, opened 2023-11-15
Populates General information section

gautierchomel:
_For easier review here is the full addition:_ 

### General information (supplement the bibliographic information)

The product details are giving precious information about usability of the book in relation with specifics needs. The following informations should always be displayed:

- File format (EPUB, PDF, MP3, Audiobook, etc.)
- Protection measure or lack thereof
- Date of publication
- Name of the publishing house
- Main language of the content

### Why this information is important for accessibility? _(note: in the code this part is a Detail, this title is the summary)_

- The file format gives a strong indication of accessibility: an MP3 format audio book will be less structured than an Audiobook; a PDF does not allows for fine typography tunning, etc.
- The protection measure may block assistive technologies such as screen readers. In addition, many specific reading devices such as DAISY readers or Braille notepads are not equipped to read encrypted files.
- The publication date gives an indication of when the files were produced. A book published before 2011 will probably be an EPUB2 file including poor semantics, and a book published in 2022 will be more likely to be correctly structured and benefit of strong semantics.
- The name of the publishing house can highlight the efforts it has made in terms of accessibility.
- The main language of the content enables readers to be sure that they will be able to read in good conditions by checking that they have a synthesised voice or the corresponding Braille table.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 UX-Guide-Metadata/bin/_tmp/
 
 UX-Guide-Metadata/UX-Guide-Metadata-principles.zip
+Gautier.md
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -149,8 +149,50 @@
 
 			<section id="general-info">
 				<h2>General information (supplement the bibliographic information)</h2>
+				<p>
+					The product details are giving precious information about usability of the book in relation with specifics needs. The following informations should always be displayed: 
+					<ul>
+						<li>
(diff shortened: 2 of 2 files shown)


gregoriopellegrino:
Suggerirei qualche modifica alla prima frase: "Although not directly related to accessibility, product details provide valuable information about the usability of the book in relation to specific needs. Therefore, the following information should always be displayed on the publication information page:"

.gitignore
clapierre, line 1: Why is there a .md file with your name on it here?

clapierre (commented):
I agree with Gregorio, other than the gautier.md file looks good.

Issue #196, opened 2023-10-30
Recommended fields for search/filtering in the user experience


rickj:
As a followup to our email conversation after TPAC, here is the sub-list of items that were identified for searching/filtering:

- Any accessible information (any of the a11y metadata fields have content)
- Conformance information (dcterm:conformsTo)
- Certification information (a11y:certifiedBy)
- Accessible Math (MathML) (accessibilityFeature = MathML)
- Supports Visual Adjustments (accessiblityFeature = displayTransformability)
- Supports nonvisual reading (accessModeSufficient = textual )

with an open question for: 'do not search/filter on hazards, but it may be critical to always be able to show the presence of claimed hazards in any search/filter return'

gautierchomel:
This aspect may be even more affected by public target of the portal / retailer. We'll need a minimal set and an extended proposal.

rickj:
I believe a minimal set should be focused on discovery, with extended set examples, or recommended details that should be displayed in the search/filter result given for specific markets.  

What I mean by 'discovery' is an answer to a user trying to find a book, and would a specific item stand alone as a search/filter term, or would it be presented in the results of that search/filter?

Examples:  Would a user say "Show me only items where _______ is true" and does this make sense for:

- show me only items that have conformance information (I think: yes)
- show me only items that have print-equivalent page numbering (I think: no, but I would want this displayed in the result)

gregoriopellegrino:
Thank you for these proposals. To me "Accessible Math (MathML)" is very important for the school sector, it seems less significant in the trade sector. What do we include in the minimal set?

rickj:
My thoughts about the context for this discussion was to provide recommendations for narrowing the list of potential metadata items (more than 70 possibilities!) that might be used to narrow down a search/filter (and avoid a very complicated user interface to support all those).  I was thinking:

- A user is coming to a site/application for discovery and is presented with a search/filter interface
- There are some 'non accessibility' metadata items that a user may input for their search/filter (author, title, ISBN, ...), and there are some 'non accessibility' metadata items that they would expect to see presented in the search/filter results (price, publisher, ...)
- The same is true for accessibility metadata items:
- There are some things they would want to use to narrow the result list down to a reasonable size (is there any a11y metadata claims at all?, has someone certified the a11y markup?, does it claim to conform to a standard?  Does it support some of the major accommodation needs (visual adjustments, non-visual reading)
[...]

PR #227, opened 2023-12-18
Adds discovering accessible content section

gautierchomel:
As per discussions at #196

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -823,17 +823,80 @@ George asks why is details used here?</detail>
 		<section id="discovering-accessible-content">
 			<h2>Discovering accessible content</h2>
 
-			<p>The guidelines for presenting accessibility metadata detailed in this document are intended to improve
-				the user experience when readers browse the catalog entry for a publication. However, accessibility
-				metadata also has a vital role to play in helping readers discover publications that are accessible for
-				them. Publication providers, such as vendors and libraries, are encouraged to create searching and
-				filtering tools that interpret accessibility metadata to aid in discovery. (A full discussion of search
-				and filtering issues is out of scope for this document, but this section provides some guidance until
-				further research can be carried out.)</p>
+			<p>The guidelines for presenting accessibility metadata detailed in this document are intended to improve the user experience when readers browse the catalogue entry for a publication. However, accessibility metadata also has a vital role to play in helping readers discover publications that are accessible to them.</p>
+			<p>Publication providers, such as vendors and libraries create searching and
+			filtering tools that interpret accessibility metadata to aid in discovery.
+			The set and variety of filters depend on the public they address and
+			the type of book they propose. </p>
+			<p> User feedback stated that in the missing of specific accessibility filters, product details like file format and measures of protections are crucial information.</p>
+			<p>Taking into consideration those realities,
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -823,17 +823,80 @@ <h4>Metadata techniques</h4>
...
+						<li>Supports Visual Adjustments</li>
+					</ul>
+				</aside>
+				<div class="note">
+					<p>In 2023, in many markets, we observe very few ebooks with accessibility metadata. 
+					   In such situation, a filter allowing for the discovery of all titles with any accessibility metadata available may be considered with interest. 
+					   We decided not to include that facet in our recommandation because of three factors: 
+					   <li>a risk of finding titles that are informing about unknown accessibility or known limitations;</li>
+					   <li>a limited utility as some accessibility metadata like the presence of TOC will cause the filter to return almost all ebooks;</li>
+					   <li>a technical implementation complexity as it would mean to grab for more than 40 information.</li>
```
clapierre, line 863: Not sure what "40 information" is.

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -823,17 +823,80 @@ <h4>Metadata techniques</h4>
...
+					<ul>
+						<li>Supports Nonvisual Reading</li>
+						<li>Supports Visual Adjustments</li>
+					</ul>
+				</aside>
+				<div class="note">
+					<p>In 2023, in many markets, we observe very few ebooks with accessibility metadata. 
+					   In such situation, a filter allowing for the discovery of all titles with any accessibility metadata available may be considered with interest. 
+					   We decided not to include that facet in our recommandation because of three factors: 
+					   <li>a risk of finding titles that are informing about unknown accessibility or known limitations;</li>
```
mattgarrish, line 861: List tags are missing around these list items.

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -823,17 +823,80 @@ <h4>Metadata techniques</h4>
...
+			the following sections are proposing a minimum and an extension set for filtering options. Mostly any specific information could be added if considered of use for the public of the platform.</p>
+
+
+			<section id="minimum-filtering-set">
+				<h3>Minimum filtering set</h3>
+				<p>
+					Reading systems, commerce, and distribution platforms will typically have specific filtering options; 
+					having uniformity for key aspects and providing guidance for a standardized approach can help 
+					the discovery process for users searching relevant titles. 
+					Accomplishing this, however, should not prevent users with specifics reading needs finding books
```
mattgarrish, line 842:
```suggestion
					Accomplishing this, however, should not prevent users with specific reading needs from finding books 
```

UX-Guide-Metadata/draft/principles/index.html
mattgarrish: It sounds like this should say something like "it would require providing more than 40 additional filtering options"

avneeshsingh: I wonder do we need to explain why we have not included this filter? This is an important discussion that we had in our meeting as well as on GitHub issue. We should focus on decision here i.e. two metadata are included in the minimum set. For the reasons why something is excluded, people can visit the GitHub issue.

PR #230, opened 2024-01-09
fixing error in respecConfig

wareid:
I noticed when I opened the new version of the UX Guide that respec wasn't formatting correctly. This should fix the issue.


mattgarrish:
You'll probably also need to add commas after Gregorio's and Gautier's entries - looks like too much got copied when adding Gautier and George to the array of editors.

mattgarrish (approved):

PR #232, opened 2024-02-07
Working draft of the new schema.org OPF accessibility metadata technique

clapierre:
Thanks Gregorio for help on this.
For group review before continuing with other sections.

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/epub-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/epub-metadata/index.html
@@ -33,860 +33,243 @@
 			};
 			// ]]>
 	  </script>
-	  <style>
-			.fake-issue {
-				padding: .5em;
-				border: .5em;
-				border-left-style: solid;
-				page-break-inside: avoid;
-				margin: 1em auto;
-			}
-
-			.fake-issue {
-				border-color: #e05252;
-				border-color: var(--issue-border);
-				background: #fbe9e9;
-				background: var(--issue-bg);
-				color: black;
-				color: var(--issue-text);
-				counter-increment: issue;
-				overflow: auto;
-			}
-
-			.fake-issue::before, .fake-issue > .marker {
-				color: #831616;
-				color: var(--issueheading-text);
-			}
-
-			.fake-issue::before, .fake-issue > .marker {
-				text-transform: uppercase;
-				padding-right: 1em;
-			}
-
-			.fake-issue a.respec-gh-label {
-				padding: 5px;
-				margin: 0 2px 0 2px;
-				font-size: 10px;
-				text-transform: none;
-				text-decoration: none;
-				font-weight: 700;
-				border-radius: 4px;
-				position: relative;
-				bottom: 2px;
-				border: none;
-				display: inline-block;
-			}
-		</style>
+		  <style>
+	  	var {
+			color: brown;
+		}
+	  	ol.condition > li {
+	  		margin-top: 1rem;
+	  		margin-left: 0;
+	  		padding-left: 0;
+	  	}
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I have a question about the or statement below. where you have:

Where you have:
3. LET real_text be the result of calling check for node on package_document, /package/metadata/meta[@property="schema:accessModeSufficient" and text()="textual"] OR the result of calling check for node on package_document, /package/metadata/meta[@property="schema:accessMode" and text()="textual"].


George asks, if the accessMode is textual, there may also be an accessMode of visual, which means there is some real text, but not necessarily all the text. This was the reason for creating the accessModeSufficient in the first place. I am not sure, but I think the only way to ensure real text is with the accessModeSufficient.

clapierre:
@GeorgeKerscher I was thinking about that as well George, 
@madeleinerothberg should also weigh in here as well but this was my thoughts.

Yes accessModeSufficient of "textual" means EVERYTHING can be viewed as text which means all images have alt text descriptions. 
So if you have an EPUB with Images (Not described) and real text for the content you would have accessMode of both "visual" and "textual" and only the accessModeSufficient of "textual,visual" but in this case for this "real_text" this is valid all text on the screen that can be seen is real, hidden text in images is not what this variable is representing.

Now I think the logic in the 2nd section still needs some work to be honest. where I think @madeleinerothberg can help with.

For example right now only when displayTransformability is present do we say "Appearance can be modified"
I think, maybe this should be if "displayTransformability" is present, OR if its "real_text".

clapierre:
Thinking about it maybe this would appear after the 
ELSE IF text_within_images OR images_of_text:
    Appearance cannot be modified for all textual content
ELSE IF real_text
    Appearance can be modified
ELSE
   Appearance modifiability not known.

This would cover if the Publisher forgot to put in displayTransformability, but had accessModeSufficient = "textual"
Or maybe real_text only is accessModeSufficient = "textual" not  OR'ed with accessMode = "textual".

Ugg my head hurts. :)

GeorgeKerscher:
I have reviewed this PR. I am going to merge it. However, I would like for us to review the logic. I believe that two items need to be true for all text to be modified. The logic says only one.

madeleinerothberg:
If accessMode consists only of "textual" (and not "textual, visual" or other combo) then that implies real_text is present because there is nothing else in the file. Note that this means that "the entire value of accessMode is "textual"" and not "value of accessMode includes "textual" among other values".

If there are both text and visuals, then you need to check if there is image description for the images to ensure the entire document can be accessed as text. accessModeSufficent is a way of encapsulating this algorithm into one element, but if we can't assume publishers will provide AMS then we can also provide the formula for deriving it.

displayTransformability is a separate decision; if the only accessMode is text it is most likely transformable. And if there are images but they aren't images of text, and we assume the user wishing to transform the text has vision, then it may be fully transformable but I don't think we can be certain if displayTransformabilty isn't set. Are we open to a statement of "Most appearance can probably be modified"?

Happy to discuss further in the meeting or here.

GeorgeKerscher:
If you go to the schema section in techniques linked from the Principles 2.0, you will see Charles section on visual adjustments.
 
Best
George

madeleinerothberg:
Yes, I wasn't very clear but my point is that case 2:

ELSE IF real_text AND NOT all_textual_content_can_be_modified:
THEN display "Appearance cannot be modified".

Might be better worded as "Most appearance can probably be modified" because if it is real_text then it should be modifiable even if the displayTransformability code is not set.

gregoriopellegrino:
We have the case of Fixed Layout EPUBs where text may be real text, but the appearance cannot be modified.

PR #235, opened 2024-02-20
Update principals fixes #199

### LINKED ISSUE #199: Overview of metadata flow
I am using this description (and image) of the metadata workflow in my conversations.  I was wondering if we wanted to add this as a note in the User Experience Guide?
-------------

1. Metadata for a given piece of content starts in a publishers ecosystem.
2. That metadata can be expressed inside of an EPUB file using the ‘EPUB Accessibility Metadata 1.1’ specification (a requirement for that specification).  Also, that metadata may be expressed in an ONIX feed (EPUB, PDF, or other formats).
3. The ‘User Experience Guide for Displaying Accessibility Metadata’ provides recommendations for how to display that metadata in a format neutral manner that is understandable and consistent.  Specifically, it provides descriptive and compact statements in 9 areas that can be used to present publication accessibility metadata in a user-friendly manner.
4. The ‘Display Techniques for EPUB Accessibility Metadata’ provides the logic for translating the machine readable metadata from EPUB files into the ‘User Experience Guide for Displaying Accessibility Metadata’.
5. The ‘Display Techniques for ONIX Accessibility Metadata’ provides the logic for translating the machine readable metadata from an ONIX feed into the ‘User Experience Guide for Displaying Accessibility Metadata’.
6. Additionally, a crosswalk between the properties defined in the EPUB Accessibility specification and the equivalents defined in the ONIX metadata standard is being maintained (https://w3c.github.io/a11y-discov-vocab/crosswalk/)

a11y metadata flow

clapierre:
Added new Publisher Metadata ecosystem from Rick Johnson.


GeorgeKerscher:
I am going to merge this and then work on the other language changes assigned to me.

There is an H4 inside the details. IMO this seems incorrect. The extended description is not part of the overall structure of the document.

GeorgeKerscher:
Rick, I put your language in before the flowchart image in that section. It is in a PR I just posted.

gregoriopellegrino:
Some comments:
- I think the diagram should have a header for itself and maybe an introduction paragraph, because right now it's inside the section "2.1 Why is this information important for accessibility?" which doesn't seem right to me
- in the chart: I don't like the division between "EPUB" and "PDF (and other) Files", can we unify naming them "digital publication"?
- in this way we can split accessibility metadata between embedded in the file (EPUB Accessibility) and metadata that travels down the supply chain (ONIX, MARC, etc.)

rickj:
Something like this?
a11y metadata flow

chrisONIX:
This is a good image. Just one thing about the block that mentions ONIX. If we put code list 196 - that excludes the important values that are conveyed elsewhere - list 81, 143 and 175 especially. Maybe it should say “accessibility metadata in ONIX” or something inclusive like that? 
thanks

rickj:
I struggled with how much detail to include in the ONIX box, and actually had code list 81 also included initially.  You comment convinces me that this is something needing change.  The [crosswalk lists elements](https://w3c.github.io/a11y-discov-vocab/crosswalk/) from codelists 03, 04, 05, 09, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19, 20, 21, 22, 24, 26, 27, 28, 29, 30, 35, 37, 75, 76, 77, 80, 81, 82, 84, 85, 86, 93, 94, 95, 96, 98, 99, 143, 144, 175.  Do we need to include all of those, or just reference the crosswalk?  

Give than [this document from EDItEUR](https://www.editeur.org/files/ONIX%203/APPNOTE%20Accessibility%20metadata%20in%20ONIX.pdf) only references codelist 196 would this work?

gregoriopellegrino:
Thank Rick for updating the chart. I would say that EPUB Accessibility 1.1 Metadata is not only schema.org, but also conformance metadata :)

gregoriopellegrino:
I think for both cases (EPUB metadata and ONIX) we may just reference generically to the accessibility metadata available in the standards, in this way we don't have to mention code lists, schema.org, ecc.

rickj:
A simplified version to evaluate:

a11y metadata flow 3

chrisONIX:
I think referencing the ONIX accessibility metadata is the best.
The list you give in your reply is the actual codes from list 196 - not the code list numbers - and then three other code lists (143, 144, 175.) The crosswalk and the ONIX documentation take care of the detail - the broad heading works here.
thanks

chrisONIX:
I like the simplified version

clapierre:
Once we have consensus, I am happy to update the image in the principals document and modify the extended description to match this new version.

gregoriopellegrino:
Fine for me

clapierre:
Currently the new version has a W3C logo in the top left corner.  Should we keep that? @avneeshsingh

rickj:
Here is a version without the logo, should the decision be to use that.

a11y metadata flow 4

PR #238, opened 2024-02-29
Implemented new changes based on 2/29 editors call

clapierre:
Using new image of Publisher metadata ecosystem from Rick.


GeorgeKerscher:
Found one typo, "". I will merge and fix this and put in an intro in this section.

PR #239, opened 2024-02-29
Added paragraph before image and fixed some typos

GeorgeKerscher:
Two typos fixed and added a paragraph as an intro before the paragraph. I do not know if this is enough of an introduction.


clapierre (approved):
Thanks for the backup.  looks great.

gregoriopellegrino:
Instead of "Schema.org" I would suggest to write "EPUB accessibility metadata", which is the common name we decided to use and which includes Schema.org metadata plus conformance metadata, pagination source reference, etc.

You might want to look at the whole document again to make sure there are no other references to Schema.org

PR #241, opened 2024-03-06
Visual adjustments updates for ONIX techniques

gregoriopellegrino:
Here are the changes made:

- simplified the technique by removing the per infer calculations, leaving only the all_textual_content_can_be_modified and is_fixed_layout options
- visually showed IDs bound to strings to be shown on screen: created red labels (formatting can be changed); they are created automatically via CSS

Can you check that assistive technologies read the labels with IDs? With VoiveOver and Chrome on Mac they seem to work.

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
@@ -217,13 +217,13 @@
 				<ol class="condition">
 					<li>
 						<span><b>IF</b> <var>all_textual_content_can_be_modified</var>:</span>
-						<span><b>THEN</b> display "<code id="visual-adjustments-1">Appearance can be modified</code>".</span>
+						<span><b>THEN</b> display <code id="visual-adjustments-1">"Appearance can be modified"</code>.</span>
 					</li>
 					<li>
 						<span><b>ELSE IF</b> <var>is_fixed_layout</var>:</span>
-						<span><b>THEN</b> display "<code id="visual-adjustments-2">Appearance cannot be modified</code>".</span>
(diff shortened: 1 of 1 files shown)


clapierre (commented):
I think the ending quotation mark should appear before the ID that is in red. Just looks like its part of the string otherwise.

EG:
```
IF all_textual_content_can_be_modified:
   THEN display "Appearance can be modified" [ID: visual-adjustments-1].
ELSE IF is_fixed_layout:
    THEN display "Appearance cannot be modified" [ID: visual-adjustments-2].
ELSE 
    display "Appearance modifiability not known" [ID: visual-adjustments-3].

```
Also I think it's best to have the last display on a separate line like shown above.

madeleinerothberg:
The new logic looks good.

I agree with Charles' suggestion about the line break.

I am unclear if the ID tags are meant to be part of the statement displayed or if they are part of the "technique" in preparation for i18n. If not intended to be displayed, then I agree with Charles about moving the closing quotation.

madeleinerothberg (approved):

mattgarrish:
I'd personally keep the "else" without a line break. There isn't a colon or "THEN" keyword or similar to justify a new line like with the others. It's just the trailing default behaviour. Otherwise, you should probably strip all the "THEN" keywords and add a colon after "ELSE" so all the statements are the same.

gregoriopellegrino:
I've made the suggested changes, other than going to head with "ELSE," we can talk about that in the editors' call.

We still need to make sure that the IDs are read correctly by JAWS and NVDA.

clapierre (approved):
Changes look good.

PR #242, opened 2024-03-07
Hazards technique for ONIX

gregoriopellegrino:
Thanks to Chris for the work.

I entered the technique of hazards. In order to handle the different permutations of possible hazards, I thought of using an array. I created a "Join array to comma list" function to concatenate the array (thanks to Matt for support).

This is a starting point. As an editor we still have some doubts, which we will dissolve in the call.

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
@@ -279,6 +279,11 @@
 						<p>If true it indicates that the <a href="https://ns.editeur.org/onix/en/143/16">code 16 of codelist 143</a> (No sound hazard warning necessary) is present in the ONIX record, otherwise if false it means that the metadata is not present.</p>
 						<p>This means there is a positive indication in the ONIX record confirming there are no sound hazards associated with this product.</p>
 					</dd>
+					<dt><var>unknown_if_contains_hazards</var></dt>
+					<dd>
+						<p>If true it indicates that the <a href="https://ns.editeur.org/onix/en/196/08">code 08 of codelist 196</a> (Unknown accessibility) is present in the ONIX record, otherwise if false it means that the metadata is not present.</p>
+						<p>This means that the product has not been assessed for hazards and there is no information about potential hazards.</p>
+					</dd>
 				</dl>
 				<h4>Variables setup</h4>
 				<ol class="condition">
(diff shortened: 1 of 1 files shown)


madeleinerothberg:
This direction makes sense. Bravo for figuring out the array and comma processing.

Should we be brainstorming the right statement for the last case (3. ELSE display "".), which I think is when we don't know if there are hazards or not? Or is that intentionally blank?

gregoriopellegrino:
> Should we be brainstorming the right statement for the last case (3. ELSE display "".), which I think is when we don't know if there are hazards or not? Or is that intentionally blank?

While working on the principles, we discussed for a long time what to show in case metadata about hazards is not present. In the end we had decided not to show anything. Hence the idea of leaving the string blank, but we can talk about that in calls.

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -173,6 +173,32 @@ <h3>Check for Node</h3>
 					<li><b>ELSE</b> return <code>False</code>.</li>
 				</ol>
 			</section>
+			<section id="join-array-to-comma-list">
```
mattgarrish, line 176: If you join an array with only one element, you get that element back, so only an array with 2 or more hazards will have commas. If you then replace the last comma with " and" you'll get the same result as this function, as it again won't alter the single-hazard case. I'd expect most languages have a "join" function, so this might be more than is needed.

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -173,6 +173,32 @@ <h3>Check for Node</h3>
...
+					<li><b>LET</b> <var>array_length</var> be the length of given <var>string_array</var>.</li>
+					<li>
+						<span><b>IF</b> <var>array_length</var> == 1:</span>
+						<span><b>THEN</b> return <var>string_array[0]</var>.</span>
+					</li>
+					<li><b>FOR</b> <var>i</var> = 0 <b>TO</b> (<var>array_length</var> - 1):
+						<ol class="condition">
+							<li><b>CONCAT</b> <var>string_array[i]</var> <b>TO</b> <var>output_string</var>.</li>
+							<li>
+								<span><b>IF</b> <var>i</var> < (<var>array_length</var> - 2):</span>
```
clapierre, line 193: Extra < present in code

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -242,7 +268,86 @@ <h3>Charts, diagrams, and formulas</h3>
...
+					<li><b>LET</b> <var>sound_hazard</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>onix</var>, <code class="xpath">/ONIXMessage/Product/DescriptiveDetail/ProductFormFeature[ProductFormFeatureType = "12" and ProductFormFeatureValue = "15"]</code>.</li>
+					<li><b>LET</b> <var>no_sound_hazards</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>onix</var>, <code class="xpath">/ONIXMessage/Product/DescriptiveDetail/ProductFormFeature[ProductFormFeatureType = "12" and ProductFormFeatureValue = "16"]</code>.</li>
+				</ol>
+				<h4>Instructions</h4>
+				<ol class="condition">
+					<li>
+						<span><b>IF</b> <var>no_hazards_or_warnings_confirmed</var> <b>OR</b> (<var>no_flashing_hazards</var> <b>AND</b> <var>no_motion_hazards</var> <b>AND</b> <var>no_sound_hazards</var>):</span>
+						<span><b>THEN</b> display <code id="hazards-no">"No hazards"</code>.</span>
```
clapierre, line 325: can we change the id to hazards-none here instead.

gregoriopellegrino:
I've made some changes, now I merge in order to move forward, but we can always go back and make changes to hazards section.

Here are the changes made:
- changed the string ID `"hazards none"` (thanks @clapierre )
- refactor of the join array steps, as suggested by @mattgarrish (and removed the function)
- added the singular string for hazards
- added the option in case the publisher specifies that it has no information about hazards
- added a note to say that metadata can be hidden in case metadata is missing

PR #249, opened 2024-03-12
ONIX techniques: first proposal for Supports nonvisual reading

gregoriopellegrino:
Thanks to Gautier and Chris for this proposal.

This is a first draft, on which we await comments.

There is a discussion among editors whether to include the possibility of there being audio or video content in the publication and then check if there are text alternatives for this content.

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
@@ -190,7 +190,65 @@
 			<section id="supports-nonvisual-reading">
 				<h3>Supports nonvisual reading</h3>
 				<p>This technique relates to <a href="https://www.w3.org/2021/09/UX-Guide-metadata-2.0/principles/#supports-nonvisual-reading">Supports nonvisual reading key information</a>.</p>
-			
+				<p>This algorithm takes the <var>onix_record_as_text</var> argument: a UTF-8 string representing the ONIX record.</p>
+				<h4>Understanding the variables</h4>
+				<dl>
+					<dt><var>all_necessary_content_textual</var></dt>
+					<dd>
+						<p>If true it indicates that the <a href="https://ns.editeur.org/onix/en/196/52">code 52 of codelist 196</a> (All non-decorative content supports reading without sight) is present in the ONIX record, otherwise if false it means that the metadata is not present.</p>
+						<p><i>All non-decorative content supports reading without sight</i> means that all contents of the digital publication necessary to use and understanding, including text, images (via their alternative descriptions), audio and video material (via their transcripts, descriptions, captions or subtitles) are fully accessible via suitable reading devices, for example text-to-speech screen readers or tactile reading devices (‘Braille displays’), and nothing in the digital publication prevents or blocks the use of alternative reading modes. The entire publication can be navigated and ‘read’ using only text rendered via sound or touch, and does not require visual perception.</p>
+					</dd>
+					<dt><var>real_text</var></dt>
+					<dd>
+						<p>If true it indicates that the <a href="https://ns.editeur.org/onix/en/81/10">code 10 of codelist 81</a> (Text) is present in the ONIX record, otherwise if false it means that the metadata is not present.</p>
+						<p><i>Text</i> means that digital publication contains "real" text (user-selectable) as its main content (or as secondary content).</p>
+					</dd>
+					<dt><var>non_textual_content_still</var></dt>
+					<dd>
+						<p>If true it indicates that at least one of the following is present in the ONIX record:</p>
+						<ul>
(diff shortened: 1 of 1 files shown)


clapierre (changes requested):
I think the variable names 
"non_textual_content_still" and "textual_alternative_still" is not clear.

Should these instead be?
non_textual_content_exists 
and 
textual_alternatives_exist_for_all_non_textual_content

gautierchomel:
> I think the variable names "non_textual_content_still" and "textual_alternative_still" is not clear.
> Should these instead be? non_textual_content_exists and textual_alternatives_exist_for_all_non_textual_content

The first intent was to include every non-textual content in the same check, including video and audio. That was a long list and would probably trigger numerous false positives (for example if there are images and video, the images have alt text but the audio has no caption). This [commit shows the list as per the first intent](https://github.com/w3c/publ-a11y/compare/main...techniques_gc#diff-70838e740c1baf42b7b6b7493b8b611254625d45b694e21ce3d8cbf04d0960aaR187).

This proposal considers only Still Images, maybe the wording can be better. Would non_textual_content_images do?

We are discussing adding additional checks for non_textual_content_motions, _non_textual_content_audio_ and the related _textual_alternative_motions_ and _textual_alternative_audio_

GeorgeKerscher:
I think the order may be wrong. We have in the principals
Supports Visual Adjustments
Supports Non-Visual reading
Conformance is the final section.

I think the previous PR made the change.

madeleinerothberg:
Are we discussing the strings to be displayed or were those decided elsewhere? If we are, I suggest editing for Instruction 3 which is when we don't have metadata stating this is a "real text" book. Since we would expect most ebooks to meet that bar, instead of saying the book "May not be fully readable in read aloud and braille" I suggest we delete the word "fully" and say it may not be readable.

For Instruction 2, where the book is in real text but has images that are not provided with alternative text, the word "fully" makes sense.

madeleinerothberg (changes requested):
See comment about Instruction 3

GeorgeKerscher:
Hi,

Where Madeleine suggests that we say for item3  It is not readable, do we mean not readable in readaloud or braille?

Also in this section, there is one typo. Where it says:
publication necessary to use and understanding, including text, images (via their alternative 

I think understanding should be understand.

madeleinerothberg:
Apologies for my laziness -- yes, I meant to propose to keep the whole sentence except the word "fully" so it would read:
"May not be readable in read aloud and braille"

gautierchomel:
@madeleinerothberg strings are in the [principles document](https://rawcdn.githack.com/w3c/publ-a11y/f7f9921cde19fd819683e9ea420482577f84088f/UX-Guide-Metadata/draft/principles/index.html#example-compact-explanations), we can propose the change there and report here.

Most ebooks with real text and few image may still be "almost fully" readable, few parts may be hard to understand if textual alternative is missing, that's why we opted for the addition of the word "fully". _may not be readable_ is not telling that most of the content is probably readable.

Exemple: a novel with a publisher logo and a photograph of the author missing alt text

GeorgeKerscher:
I believe we are defining the strings in the principles section. We are referencing the compact presentation in the techniques.

Here is what the principles section says right now:
EXAMPLE 4: Compact explanations
• Readable in read aloud and braille.
• May not be fully readable in read aloud and braille.
• Not fully readable in read aloud and braille.

Again, there are subtile  differences in the last two. I am not sure if people will get the differences.

Might we say: • Portions may not be readable in read aloud and braille.
and for the final case: Portions will not be readable in read aloud and braille

GeorgeKerscher:
This section still needs the order fixed to align with the Principles and the Schema section.

Non visual reading should be first.

madeleinerothberg:
> @madeleinerothberg strings are in the [principles document](https://rawcdn.githack.com/w3c/publ-a11y/f7f9921cde19fd819683e9ea420482577f84088f/UX-Guide-Metadata/draft/principles/index.html#example-compact-explanations), we can propose the change there and report here.

OK, I'll try to figure out how to open an issues there, if after reading this comment others agree that it is needed. I may be wrong.
 
> Most ebooks with real text and few image may still be "almost fully" readable, few parts may be hard to understand if textual alternative is missing, that's why we opted for the addition of the word "fully". _may not be readable_ is not telling that most of the content is probably readable.
> 
> Exemple: a novel with a publisher logo and a photograph of the author missing alt text

I agree -- but wouldn't these books will fall under Instruction 2 -- real_text AND non_textual_content_still AND NOT textual_alternative_still.

I am discussing the ones that fail instruction 2 and fall through to instruction 3. Books with no real_text are not "maybe not fully readable" but instead "may not be readable at all" because they have no assertion of real text.

avneeshsingh:
If we are OK with the technique, and the improvements are confined to text in principles, then we should merge this pull request and update it again when we finalize the text for principles.

PR #258, opened 2024-04-04
Proposal for ONIX Pre-recorded audio

gregoriopellegrino:
Thanks to @gautierchomel for the draft!

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
@@ -297,7 +297,58 @@
 			<section id="pre-recorded-audio">
 				<h3>Pre-recorded audio</h3>
 				<p>This technique relates to <a href="https://www.w3.org/2021/09/UX-Guide-metadata-2.0/principles/#pre-recorded-audio">Pre-recorded audio key information</a>.</p>
-
+				<p>This algorithm takes the <var>onix_record_as_text</var> argument: a UTF-8 string representing the ONIX record.</p>
+				
+				<h4>Understanding the variables</h4>
+				<dl>
+					<dt><var>audiobook</var></dt>
+					<dd>
+						<p>If true it indicates that the <a href="https://ns.editeur.org/onix/en/81/01">code 01 of codelist 81</a> (Audiobook) is present in the ONIX record, otherwise if false it means that the metadata is not present.</p>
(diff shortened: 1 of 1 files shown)


clapierre:
Gregorio, there are 4 variables defined and a lot of AND / NOT Else if combinations I am wondering if there are missing combinations we need to ensure gets addressed otherwise they will all default to your Else clause.

gregoriopellegrino:
> Gregorio, there are 4 variables defined and a lot of AND / NOT Else if combinations I am wondering if there are missing combinations we need to ensure gets addressed otherwise they will all default to your Else clause. Maybe those wouldn't ever occur together, or if they did are in conflict with each other, not sure unless you map out the entire matrix of possibilities, which maybe you have already done?

On this I turn the question to @gautierchomel

gautierchomel:
@clapierre I think we are covering almost all cases here, but I may be missing something, I'm not very familiar with audiobooks. 

There's maybe a discussion to have about usecases for Complementary Audio and Text, like code 21 of codelist 81 (Partial performance – spoken word) or code 22 of codelist 81 (Additional audio content not part of main content). I understand this pre recorded audio value as "_Main content is audio"_ while those are more like _"there are some bits of audio inside"_ and therefore I would prefer to address them in Support non visual reading section.

gautierchomel:
I added a second possible trigger for _synchronised_pre_recorded_audio_ as I realised that the Product description (code A305, list 175)  is as relevant as the accessibility specific one (code 20, list 196). 
This is: code A305 of list 175  Synchronised audio (Audio synchronised to text within an e-publication, for example, an EPUB3 with audio overlay. Synchronisation at least at paragraph level, and covering the full content)

gautierchomel:
Relooking at the Display principle, I see that we have written _... accompanies text (embedded audio and video clips) ..._ ([3.4 Pre-recorded audio](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/draft/principles/?updated#pre-recorded-audio))

So I'm adding those use cases. 

Thanks @clapierre for triggering the alarm.

gautierchomel:
Here is the [preview](https://rawcdn.githack.com/w3c/publ-a11y/b65e4f0c86afc1f3d22ea5a345d044cacfb7324d/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html#pre-recorded-audio)

PR #259, opened 2024-04-10
added new technique for non visual reading

clapierre:
Also fixed a few other issues found missing ending }, duplicate section, and removed the complexity of text()= to improve readability.
Definitely needs some extra review here as I am not 100% confident this is completely correct.
Especially since ONIX has "real_text" variable I didn't include here as I wasn't sure how it mapped.

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/epub-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/epub-metadata/index.html
@@ -58,9 +58,10 @@
 			margin-left: 0.2em;
 			padding: 0.1em;
 		}
-		code[id]::after {
-			content: "[ID: " attr(id) "]";
- 		/*html::before { content: "<html>"; }*/
+              code[id]::after {
+                  content: "[ID: " attr(id) "]";
+                  /*html::before { content: "<html>"; } */
+              }
 	  </style>
 	</head>
 	<body>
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
Hi Charles,
the structure looks good to me, but the XPATHs don't seem correct, for example `/package/metadata/meta[@property="schema:accessibilityFeature" = displayTransformability]` is not a valid XPATH.

I recommend that you test XPATHs, for example you can use the free site http://xpather.com/

clapierre:
Ok, I was trying to make it simpler to read, I will revert back to the original way with the text()=... as well then.

clapierre:
Is this now correct @gregoriopellegrino?  

/package/metadata/meta[
@property="schema:accessibilityFeature" and text()="alternativeText" or
@property="schema:accessibilityFeature" and text()="longDescription" or
@property="schema:accessibilityFeature" and text()="describedMath"].

Or do I need separate xpath's for each "or" statement?

like this?
/package/metadata/meta[@property="schema:accessibilityFeature" and text()="alternativeText"] or
/package/metadata/meta[@property="schema:accessibilityFeature" and text()="longDescription"] or
/package/metadata/meta[@property="schema:accessibilityFeature" and text()="describedMath"].

clapierre:
Actually it looks like I still have it wrong.  Seems like this is the correct approach for xpath according to AI Claude.

/package/metadata/meta[@Property="schema:accessibilityFeature" and (@Value="alternativeText" or @Value="longDescription" or @Value="describedMath")]

clapierre:
fixed the xpaths for these more complicated logic scenarios. @gregoriopellegrino how are these now?

GeorgeKerscher:
I reviewed PR 259, and I cannot find any problems. I compared the section with its ONIX counterpart. I will merge.

PR #261, opened 2024-04-22
First draft for Accessibility summary or addendum

gautierchomel:
Variables setup are missing, @gregoriopellegrino proposed a different method to set up the language of the Summary or Addendum.

### DIFF
--- a/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
+++ b/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
@@ -95,7 +95,7 @@
 				It is important to use ONIX metadata as a complement to the accessibility data embedded within the publication itself, if describing accessible books, books with full audio and related products for the global book supply chain.
 				An ONIX file can be used to display accessibility information in advance of publication or when you do not have access to the metadata in the digital file itself.
 				Some accessibility information may only be available when you have access to the file itself.
-				If you are unfamiliar with ONIX, <b>THEN</b> there is more documentation available from <a href="https://www.editeur.org/83/Overview/">EDItEUR.org</a>.
+				If you are unfamiliar with ONIX, then there is more documentation available from <a href="https://www.editeur.org/83/Overview/">EDItEUR.org</a>.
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -580,9 +580,49 @@ <h4>Instructions</h4>
 			</section>
 			
 			<section id="accessibility-summary">
-				<h3>Accessibility summary</h3>
+				<h3>Accessibility summary or addendum</h3>
```
gregoriopellegrino, line 583: I would only leave "Accessibility summary" to be in line with the Principles

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -580,9 +580,49 @@ <h4>Instructions</h4>
...
+					</dd>
+				</dl>
+				<h4>Variables setup</h4>
+				<ol class="condition">
+					{{/*  <li><b>LET</b> <var>onix</var> be the result of calling <a href="#pre-processing">pre processing</a> given <var>onix_record_as_text</var>.</li>  */}}
+				</ol>
+				<h4>Instructions</h4>
+				<ol class="condition">
+					<li><span><b>IF</b> <var>accessibility_addendum</var><b>AND</b> <var>accessibility_summary</var>
+						<b>THEN</b> display the content of <var>accessibility_addendum</var> in a Span with language_of_text as an argument</li>
```
gregoriopellegrino, line 619: I would not use "Span" because it is only related to HTML, I would write "display the content of accessibility_addendum with language_of_text as language"

gregoriopellegrino:
I think the IF-THEN block can be simplified, I try to send a proposal

gregoriopellegrino:
I have pushed a proposal with some modifications and XPATHs

clapierre (approved):
Like todays improvements. 
Looks good

PR #267, opened 2024-05-06
Updated principles around unknown discussion

GeorgeKerscher:
I maded the changes from the  issue 262 discussion.

First, I updated the note about unknown and relevant information and made it more explicit.

Then I fixed the missing detailed information that was missing the credential and report. This was a unordered list and I changed it to a paragraph to match the others.

Finally, I add an unknown example where the conformance metadata was missing.


clapierre (approved):
Looks good George, you could capitalize the "Four examples" when starting that sentence. other than that looks good.

PR #272, opened 2024-05-17
Update pre recorded audio

clapierre:
Note this is slightly different than the ONIX version, and will require some careful review / discussion.


GeorgeKerscher:
I have a few questions:

First,  in the vvriable setup it says:
/package/metadata/meta[@property="schema:accessModeSufficient" and text()="auditory"].

I don't know if the and text () is correct. Somebody should check. This appears in several places here.

Second, regarding accessibilityFeature="ttsMarkup" , I understand that this has the potential to improve TTS pronunciations , but does it belong in the  pre-recorded audio section? It may be that the synchronized audio, if generated by TTS would probably be better, but   should we say anything in this section? I don't think we know if the audio is a human  or TTS.

gregoriopellegrino:
> I don't know if the and text () is correct. Somebody should check. This appears in several places here.

That is correct. In the XPATH language, it is the way to identify the value contained in an XML tag, so it is the way to find the XML element with a certain value.

clapierre:
@GeorgeKerscher, I was mistaken as I thought the ONIX code's 20 and 21 was referring to Code List 196 which does align with ttsMarkup as we have in our Crosswalk document. I agree that the doesn't belong for pre-recorded audio, so I will remove this.

clapierre:
Ok, I made that change @GeorgeKerscher and removed ttsMarkup from the EPUB pre-recorded audio technique.

clapierre:
With the changes made, can we merge this?  @GeorgeKerscher @gregoriopellegrino

PR #274, opened 2024-05-20
added EPUB Technique for Navigation Metadata

clapierre:


GeorgeKerscher:
I have several comments:
1. It says:
This algorithm takes the onix_record_as_text argument: a UTF-8 string representing the ONIX record.

not ONIX.
2. It says: This means that there is a Table of Contents in the publication that allows direct (eg hyperlinked) access to all levels of text organization above individual paragraphs (eg to all sections and subsections) and to all tables, figures, illustrations etc.

George: I think this goes a bit too far. Using All is the first problem. I would think it provides direct access to the structures of the title that the publisher has selected. It may be that headings down to level 2 or 3 are provided, but the publisher may not go down to the lowest level.

Also, all tables and illistrations seems to be too much.

clapierre:
@GeorgeKerscher Fixed the ONIX reference, nice catch.
As for issue 2 I just took the text which @gregoriopellegrino  wrote in the ONIX technique so lets come up with what this should be and then both Gregorio and I can update our documents.

I agree that how does having a table of Contents present in the metadata gives access to the tables, figures, and illustration's not sure if having this implies direct access to those unless there are separate navigation list of figures, list of tables etc. which I don't believe we have any metadata for.  At least we don't in EPUB schema.org metadata.

clapierre:
lets change this to 

The resource includes a table of contents that provides links to the major sections of the content.

clapierre:
Added final change with improved TOC definition added as discussed.  Please review @GeorgeKerscher and @gregoriopellegrino

GeorgeKerscher:
I think this looks great. I don't know about the array, but assuming it works, I totally approve.

There are conflicts, so Charles should resolve. After that, I think it is good to merge.

clapierre:
conflict was just some initial blank space / tabs

PR #281, opened 2024-05-21
Added note with text from comment in 271

GeorgeKerscher:
I took the text from the comment in 271. I put it at the end of the section that links to the techniques. The text says:
Publishers update their ONIX records as needed. We expect unknown accessibility metadata will be provided as more information becomes known. For this reason, implementors should be prepared to update the accessibility metadata as new ONIX feeds are made available.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -196,6 +196,9 @@
 						<p><a href="../techniques/onix-metadata/index.html">ONIX Accessibility Metadata</a></p>
 					</li>
 				</ul>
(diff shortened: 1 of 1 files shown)


clapierre (commented):
Does this statement really make sense?

We expect unknown accessibility metadata will be provided as more information becomes known.

Shouldn't it be something more like:

We expect "unknown" accessibility metadata may be initially provided but may change as more information becomes known.

GeorgeKerscher:
The latest commit adds the language Charles has suggested. I think it is much better.

clapierre (approved):

PR #285, opened 2024-05-21
added new EPUB Charts, Diagrams, and Formulas Technique

clapierre:
I approve this PR. However, I do see an issue that I do not how to address with our current metadata. Chem on visual indicates that there is Chemistry present as a visual. It could be made accessible through ChemML, but I do not know how widely it is used. It could also be made accessible with a long description and now that MathML supports Chemistry, that could also make it accessible. So, for now I think we are good, but should I submit an issue about this. There is also the Community Group Chemistry on the Web and in Publishing we could turn to for guidance.


clapierre:
@gregoriopellegrino Do you approve? I think @GeorgeKerscher approved but for some reason GitHub has my name attached to George's comments.  Very odd.

gregoriopellegrino (approved):

PR #289, opened 2024-05-23
Correct Heading orders for Localization and Implementations

gautierchomel:
Those titles where h3, so under <h2>Discovering accessible content</h2>. This change makes them independant.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -542,7 +542,6 @@
 						
 
 					</aside>
-				</se
 				</section>
 
 				<section id="conf-techniques">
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I see that you mad the two sections as an h2, but when I look at it locally, they are showing up as h3 and I don't know why. Is there a problem with needing to close a section first, or might it be a respec issue?

GeorgeKerscher:
I made a commit that fixes the heading levels. There was a missing opening and closing section tag. I think I  got this right.

mattgarrish:
There was a broken closing section tag at line 545 and extra closing section tags at lines 899 and 1067. Validating in oxygen the markup is now valid. It's probably long overdue to run a indent and line length format on the file, though.

PR #290, opened 2024-05-23
updated accessibilitySummary Technique

clapierre:


gregoriopellegrino:
It seems to me that the explanation of the language_of_text variable is incorrect: it is the language metadata of the EPUB, right?

clapierre:
I am getting the `<dc:language>` which according to the spec.

EPUB 3.x
The first `<dc:language>` element must be considered the primary language of the publication.

`/package/metadata/dc:language/text()`

What are you suggesting @gregoriopellegrino ?

gregoriopellegrino:
You should use:

``
/package/metadata/dc:language[1]/text()
``

gregoriopellegrino:
Not all EPUBs have the lang attribute on the OPF element, so the language of the book is a fallback (this is how we defined it in ONIX).

clapierre:
@gregoriopellegrino  thanks and updated now.

GeorgeKerscher:
Four items. 

1. Where it says: Being human-written, the accessibility summary and addendum will appear in the original language of 
George asks: What is the and addendum? 

2. Where it says:
This means there is a human-written text containing a short explanatory summary of the accessibility of the product or the URL of a web page comprising such a summary. Summarizes the already existent information and may add information that the publisher could not express with the other codes.

George: Remove "a" add the word It, removed already, and changed existent  to existing.

This means there is human-written text containing a short explanatory summary of the accessibility of the product or the URL of a web page comprising such a summary. It summarizes the existing information and may add information that the publisher could not express with the other codes.

3. Spelling: nearest anchestor -- should be  ancestor

4. It says: This indicates the main language of the content and therefore the most probable language of accessibility summary.
George: Insert "the" before  accessibility summary.

gregoriopellegrino:
> 1. Where it says: Being human-written, the accessibility summary and addendum will appear in the original language of
>    George asks: What is the and addendum?

It is a metadata in ONIX: in order not to change the meaning of the accessibility summary metadata, they left the metadata as is and added a code for the new accessibility summary role (called an addendum) in which to put what is not expressed with machine readable data.

avneeshsingh:
Can EPUB metadata differentiate  between "AccessibilitySummary" and "addendum"?
In ONIX there is a separate code for providing addendum, but it is not the case for schema.org/EPUB metadata. So, should we use this term in EPUB metadata techniques?

gregoriopellegrino:
I don't think it's necessary: in ONIX, for backward compatibility issues EditEUR couldn't change the meaning of the accessibility summary. In EPUB metadata we changed the meaning of accessibility summary in EPUB Accessibility 1.1, so I don't think it's necessary to add a field. The result for the end user is the same.

gautierchomel:
> Can EPUB metadata differentiate between "AccessibilitySummary" and "addendum"?
> In ONIX there is a separate code for providing addendum, but it is not the case for schema.org/EPUB metadata. So, should we use this term in EPUB metadata techniques?

The ONIX approach is more robust, so it is possible to separate the old approach (accessibility summary) from the new one (accessibility addendum). In the future, we can prioritize the Addendum over the Summary and eventually deprecate the use of the second (at least not recommending displaying it as it is a repetition of the machine-readable metadata).

But that means backpedaling on the changes already made.

clapierre:
Are we ready to merge now? @GeorgeKerscher @gautierchomel @madeleinerothberg 
Here is a [link to the Rendered view](https://raw.githack.com/w3c/publ-a11y/EPUB-Technique-A11y-Summary/UX-Guide-Metadata/draft/techniques/epub-metadata/index.html#accessibility-summary)

gautierchomel:
That's detail, but to avoid confusion, I would suggest to delete:

- the words "and addendum" in the first note
- the second note which recommends to prioritize addendum over summary

GeorgeKerscher:
I don't think we should reference an an addendum in  .

clapierre:
Hi @gautierchomel @GeorgeKerscher and @gregoriopellegrino I have removed the addendum references and tweaked the note regarding what publishers should put in the summary to avoid being redundant.

gregoriopellegrino (approved):
Fine for me, thank you

clapierre:
As decided in the Editors call, this is now good to be merged.

PR #298, opened 2024-06-06
#280 proposal: The exemption reason should not be displayed unless the publisher opts in for it.

### LINKED ISSUE #280: Do we want to display Reason why the publication does not claim to meet the standards
At [Detailed conformance information](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/draft/principles/#conf-aai) we suggest to display: 

> Reason why the publication does not claim to meet the standards. For some jurisdictions, it may be important to indicate why a publication is not compliant with the standards, such as if the company is too small and therefore not required to, or if the cost to produce the accessible version is too high. Showing this information may be important if required by local legislation, otherwise it can be omitted.

That's materialised by [EXAMPLE 7](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/draft/principles/#example-unknown-accessibility-conformance-statement-with-detailed-conformance-information): 

> Unknown accessibility conformance statement with detailed conformance Information
> Conformance to accepted standards for accessibility of this publication cannot be determined.
> 
> Detailed Conformance Information:
> 
> The creator of the publication is not required to publish in an accessible format because it is a micro-enterprise (e.g., enterprises employing fewer than 10 people and with annual turnover or balance sheet total not exceeding €2 million).

While this information is of use for control authorities, 

1. is this information valuable for the end user? and 
2. do publishers want their ebooks to be displayed with negative statements while in fact they will still probably have most of the accessibility features?

A side effect to consider is that an ebook can be made fully accessible because it's only text with a cover and logo as images, but to avoid burden and cost, the publisher indicates being exempted for micro-enterprise.  Some people may interpret this ebook as not accessible enough for them, while it is not the case.

gautierchomel:
The exemption reason should not be displayed unless the publisher opts in for it.

This is a proposal to resolve #280 by clarifying that the machine-readable exemption metadata is for authority control and eventually filtering options but should never be displayed as it exposes publishing house privacy (like annual turnover).

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -400,11 +400,7 @@
 					placed immediately after the conformance statement.</p>
 
 				<aside class="note">
-					<p>In cases where the publication's compliance with accessibility standards is unknown, or it has
-						known accessibility problems, it may be important for local legislation (e.g., the European
-						Accessibility Act) to indicate why the publication is not accessible. This information could be
-						contained in the Accessibility Summary, or as machine readable metadata, in which case the
-						information should be given in the detailed accessibility information section.</p>
+					<p>Because it may reveal private information related to the publishing company, reasons of exemptions to legislation should not be displayed unless a local law obliges to do so. A publisher willing to have this information provided to the end user can provide it thru the accessibility addendum or summary</p>
 				</aside>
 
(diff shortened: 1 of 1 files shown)


gautierchomel:
Closing this PR as a consensus has been found around displaying one general statement for any exemption. More feedback is expected at #350

PR #328, opened 2024-06-06
Removed the personal pronouns in background section

GeorgeKerscher:
Matt suggested that we remove the personal pronouns in the guidelines. While this is not a spec, it still should be avoided. This is in the background section.


clapierre (approved):

mattgarrish (approved):
LGTM

PR #339, opened 2024-06-13
Made changes in conformance section

GeorgeKerscher:
Added a note that explains the exemption problem. Also, changed a should to a may where we talk about showing the details.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -472,12 +472,6 @@
 						<dt>Certifier's Report</dt>
 						<dd>If a link to a report is provided, this may be of interest.</dd>
 
-						<dt>Reason why the publication does not claim to meet the standards</dt>
-						<dd>For some jurisdictions, it may be important to indicate why a publication is not compliant
-							with the standards, such as if the company is too small and therefore not required to, or if
-							the cost to produce the accessible version is too high. Showing this information may be
-							important if required by local legislation, otherwise it can be omitted.</dd>
-					</dl>
 				</section>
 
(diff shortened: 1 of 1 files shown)


gautierchomel:
I would like to add clarity on the consensus at this point:

1. If we do not suggest displaying it, publishers who want to give users exemption information can use the Summary to do so.
2. If we suggest displaying it, publishers who don’t want it to be displayed must not use that metadata, which has other consequences for control, filtering, and data collection. 

Option one is less likely to disturb or create a threat, so it is the most convenient one for consensus.

Also, an attention point raised regarding the confusion between Conformity to a standard, exemption to legal provisions, and user needs for information on accessibility features. 

To address that, the proposal is:

1. We keep the conformity section for conformance only (ONIX 196, 01 to 05 and 80 to 99; EPUB conformsTo and affiliated), 
2. The value “The publication does not include a conformance claim” is triggered when no conformity code is provided. 
3. The wording “The accessibility of this publication is unknown” Is changed to The conformity to a standard of this publication is unknown.
4. We delete the values “This publication is known to have accessibility limitations” as both ONIX 09 and Schema Accessibility Feature None are not about conformity. The absence of feature is already available to the end user as Non Visual Reading and Display Transformability are to be informed if Not possible or Unknown. 
5. We include nothing that needs a check related to exemption.

We can open separate issues to discuss:
1. What and how legal information may be displayed?
2. What and how known accessibility limitations can be informed and displayed?

gautierchomel:
> We should avoid changing the PR without discussions with the original creator

True, I wanted to highlight the changes proposed but should not have been done by rewriting previous work. This should be fixed now.

PR #340, opened 2024-06-17
Distinguishing conformity / Legal provision / Accessibility (for editor call discussion)

gautierchomel:
This is a follow-up to the discussion that started with the display of exemptions. It proposes to keep the conformance section straight about Conformance. Aside from this, we can open separate issues to discuss:

1. What and how legal information may be displayed?
2. What and how known accessibility limitations can be informed and displayed?

Regarding the different options for Displaying the reason for the exemption:

1. If we do not suggest displaying it, publishers who want to give users exemption information can use the Summary to do so.
2. If we suggest displaying it, publishers who don’t want it to be displayed must not use that metadata, which has other consequences for control, filtering, and data collection.

Option one is less likely to disturb or create a threat, so I feel it is the most convenient one for consensus.

Also, an attention point was raised regarding the confusion between "Conformity to a standard", "exemption to legal obligations", and "user needs for information on accessibility features".

To address that, my proposal is:

1. We keep the conformity section for conformance only (ONIX 196, 01 to 05 and 80 to 99; EPUB conformsTo and affiliated),
2. The value “The publication does not include a conformance claim” is triggered when no conformity code is provided.
3. The wording “The accessibility of this publication is unknown” Is changed to "The conformity of this publication to a standard is unknown".
4. We delete the values “This publication is known to have accessibility limitations” as both ONIX 09 and Schema Accessibility Feature None are not about conformity. The absence of feature is already available to the end user as Non Visual Reading and Display Transformability are to be informed if Not possible or Unknown.
5. We include nothing that needs a check related to exemption.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -472,12 +472,6 @@
 						<dt>Certifier's Report</dt>
 						<dd>If a link to a report is provided, this may be of interest.</dd>
 
-						<dt>Reason why the publication does not claim to meet the standards</dt>
-						<dd>For some jurisdictions, it may be important to indicate why a publication is not compliant
-							with the standards, such as if the company is too small and therefore not required to, or if
-							the cost to produce the accessible version is too high. Showing this information may be
-							important if required by local legislation, otherwise it can be omitted.</dd>
-					</dl>
 				</section>
 
 				<section id="conf-examples">
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -404,14 +404,6 @@
 					publication. The certifying organization should be identified along with their credentials and
 					placed immediately after the conformance statement.</p>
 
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
Thank you, let's talk about it on Thursday in the call.

clapierre:
What is the time of this call?  For some reason I have the same call at 7AM and at 8AM PDT in my calendar.


Thanks

Charles

EOM



________________________________

GeorgeKerscher:
I don’t have anything in my calendar.
 
Best
George

GeorgeKerscher:
I think this is a good approach. I will need to remove the statement that there are four examples. That will be for another pr.

UX-Guide-Metadata/draft/principles/index.html
GeorgeKerscher, line 1: I think this looks good. I will need to change the statement that there are four examples, but that will be for another PR.

gregoriopellegrino:
Before taking actions, I propose to discuss this in tomorrow's call.

PR #342, opened 2024-06-22
Conformance section fixed

GeorgeKerscher:
This pr changes the statement in conformance from four to three. It also adds the possibility of including the following statements:
• The appearance cannot be modified. This statement was already in the compact.
• Content will not be available in generated read aloud speech and electronic braille.
• Not readable in read aloud and braille.

The above would be triggered by ONIX 09 or if accessibilityFeatures is set to none.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -292,6 +292,7 @@
 							<li>Text and page layout cannot be modified. The reading experience is close to a print
 								version. Reading systems can still provide zooming options.</li>
 							<li>Ability to modify the appearance is not known.</li>
+<li>The appearance cannot be modified.</li>
 						</ul>
 					</aside>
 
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -471,9 +471,9 @@ <h4>Detailed conformance information</h4>
 				<section id="conf-examples">
 					<h4>Examples</h4>
 
-					<p>Four  examples are provided for the conformance statement, one shows a statement that claims to
+					<p>Three examples are provided for the conformance statement, one shows a statement that claims to
 						meet recommended accessibility standards and a second that claims to meet the minimum level. The
-						third and fourth show a publication with unknown accessibility.</p>
+						third   shows a publication with unknown accessibility.</p>
```
gautierchomel, line 480: The third example should be "unknown conformity"

gautierchomel (changes requested):
The third example should be "unknown conformity" instead of "unknown accessibility"

PR #343, opened 2024-06-22
Aligned hazards language with techniques

GeorgeKerscher:
The hazards section was using different language  than the techniques. This Pr aligns the language.


GeorgeKerscher:
I tried to do this whol PR in VS Code. I think it went well, but the selection of reviewers was tricky. Hopefully you get the request, and only once, grin.

clapierre (approved):
worked great George, glad its working with VS.

gautierchomel (approved):

PR #345, opened 2024-06-28
added claims, declaration, and other suggestions

GeorgeKerscher:
suggestions from Gregorio were inserted. I also went through the introduction and first sections and used the word "claims" and "declare" in several places. I did not create a heading talking about claims, and I did not create a note to describe the destinction.

In the conformance the word claims is used. Also, once we have the implementation samples that use the claims or declarations termonology, I think it should be clear.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -109,7 +109,7 @@
 	<body>
 		<section id="abstract">
 
-			<p>The accessibility of a publication is useful to know regardless of a person's abilities, as features such as the ability to make visual adjustments make for a better reading experience for everybody. This document proposes a shared framework for presenting publication accessibility metadata in a user-friendly manner — to offer the information to end users in a way that is easy to understand regardless of their technical knowledge and is consistent across different publications and different digital catalogs. </p>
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -160,7 +160,7 @@ <h2>General overview</h2>
 				version of the EPUB Accessibility specification, so a more in-depth analysis in the <a
 					href="#accessibility-summary">Accessibility summary</a> section is recommended.</p>
 
-			<p>This document offers guidance on how to aggregate and display metadata to end users; these are not strict
+			<p>This document offers guidance on how to aggregate and display claims Inherent  in metadata to end users; these are not strict
```
clapierre, line 163: inherent should be lowercase

clapierre (approved):

gregoriopellegrino:
Two things:
- do we want to suggest that implementers call the section where they will present this data "Accessibility declaration"?
- the conformance section has the word "claim" in multiple places, do we want to leave it in?

GeorgeKerscher:
We do not yet have our implementation section. I think these will have headings of Accessibility Claims or Declarations. If we want to make this recommendation in the Principles in some other place, I am open to that. That would be for another PR.

PR #346, opened 2024-07-04
Adding text of known limitations on accessibility to acc. summary

gregoriopellegrino:
For ONIX techniques: adding human-written text containing details of and reasons for limitations on accessibility (if present) to accessibility summary section.


GeorgeKerscher:
Hi,

I do not know how to put in inline comments or suggestions. I will therefore put the suggestions here.

Where it says: Therefore it is necessary to take care of setting up 
I suggest changing the "of" to "to".

In several places it says anchestor and the correct spelling is ancestor.
For example: of the nearest anchestor.

Where it says: Contains precisions that the publisher could not 
I suggest: It contains precise information that the publisher

clapierre (approved):
other than George's comments this looks good.

gregoriopellegrino:
I have incorporated George's suggestions, thank you!

PR #349, opened 2024-07-10
Claims and legal

GeorgeKerscher:
Mad the heading for claims and declaration. Removed the word claims from thedescriptive and compact statements. I left the word claims in various places for readability and clarity.

I added the legal considerations section. Added the disclamer. Added a link to our issue tracker and we are reviewing this. Add the exemption and exception information. Because if this metadata is present, there is only one statement to be made, so the examples have one item for descriptive and another for compact.
Spell checked and checked the heading structure.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -798,11 +798,11 @@
 <div class="note">Disclaimer: nothing here may be construed as legal advice supplied by W3C or any the groups or individuals who have contributed to this document.
 Always get legal advice from in-house or your retained counsel to assess legal compliance and risk.</div>
 
-<p>In some jurisdictions  publishers may have an exemption from the provision of accessible publications, including the provision of accessibility metadata. One  example of this is  from the EAA when the publisher    is a micro-enterprise (i.e., enterprises employing fewer than 10 people and with annual
-turnover or balance sheet total not exceeding 2 million euro).</p>
-<p> Other legal considerations may surround exceptions, where jurisdiction  will not require the title to be accessible if it involves  a fundamental alteration of the content, or if making it accessible would place   a disproportionate burden on the publisher.</p>
+<p>In some jurisdictions publishers may be able to claim an exemption from the provision of accessible publications, including the provision of accessibility metadata. This should always be subject to clarification by legal counsel for each jurisdiction.
+	One example of this, at time of writing, is from the EAA when the publisher is a micro-enterprise (i.e., enterprises employing fewer than 10 people and with annual turnover or balance sheet total not exceeding 2 million euro).</p>
+<p>Other legal considerations currently included in the EAA may include exceptions on individual versions of a title, where jurisdiction will not require the title to be accessible if it involves a fundamental alteration of the content, or if making it accessible would place a disproportionate burden on the publisher. (This may vary in other jurisdictions).</p>
 
-<p > When      the publisher is claiming an exemption or an exception  they  may not  want their reason  for not providing accessibility information to be  disclosed to end users. The metadata may  be  delivered for Business-to-business purposes. Nevertheless, the objective is to provide as much accessibility information as is possible."</p>
+<p>Publishers may need to include information about an exemption or exception in metadata for legal or clarity reasons, either to bodies that enforce legislation or for other business to business communication. However this is not information that needs to be displayed on public sites as it does not mean anything to most consumers and could lead to misunderstandings.Nevertheless, the objective is to provide as much accessibility information as is possible."</p>
 
             <section id="legal-examples">
 					<h2>Examples</h2>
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
The work seems to me to be an excellent starting point. Below are some of my comments:
- for the note linked to the issue I suggest using ReSpec markup so that it is directly linked to the GitHub issue: https://respec.org/docs/#issue, in our case it should be `<div class="issue" data-number="280"></div>`
- in the "Legal considerations" sentence, the sentence "When the publisher is claiming an exemption or an exception they may not want their reason for not providing accessibility information to be made public..." doesn't sound very convincing to me, it seems like an unclear statement, besides the fact that an output suggestion is indicated, which is then given below in the examples
- in the "Legal considerations" sentence, I would add a paragraph about how useful this information can be for those in the accessible digital publishing supply chain and for use cases such as procurement
- in the "Legal considerations" sentence, for the time being I would not put examples, I would leave a placeholder, waiting for consensus on possible texts to be included

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -781,7 +787,46 @@ <h4>Metadata techniques</h4>
 					</ul>
 				</section>
 			</section>
+<section id="legal-considerations">
+<h3>Legal considerations</h3>
+<div class="note">
+					<p>This key information can be hidden if metadata is missing.</p>
```
clapierre, line 793: "missing" here seems to imply that the publisher forgot to include this information.  Maybe "not present"?

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -781,7 +787,46 @@ <h4>Metadata techniques</h4>
...
+</aside>
+</section>
+
+						
+<section id="legal-techniques">
+					<h4>Metadata techniques</h4>
 
+					<ul>
+						<li><a href="../techniques/epub-metadata/index.html">EPUB Accessibility: legal considerations</a></li>
+						<li><a href="../techniques/onix-metadata/index.html">ONIX: legal considerations</a></li>
```
clapierre, line 825: Should we add perhaps #legal-considerations to the end of these links?

GeorgeKerscher:
I made the changes suggested by Charles, Gregorio, and what was discussed on the call.

I created a new issue for publisher feedback which is in the Legal considerations section. It has a link to the prior issue. I created a link in the implementation section requesting organizations and companies who want to volunteer to put in their contact information.

gregoriopellegrino:
Hi,
it seems to me that it is taking shape. I think that in yesterday's call we found consensus to include among the examples the phrase "This ebook benefits from a legal exemption to accessibility obligations".

Then there is the sentence:

> When the publisher is claiming an exemption or an exception they may not want their reason for not providing accessibility information to be disclosed to end users. The metadata may be delivered for Business-to-business purposes. Nevertheless, the objective is to provide as much accessibility information as is possible.

In my opinion it is not very clear. Do you think it can be rewritten or removed? @gautierchomel do you have any proposals?

Also the next suggestion:

> It may be advisable for implementors to instead provide guidance to the end user by including a user-friendly statement such as, "This title may not yet be fully accessible to all readers. Check the list of accessibility features to determine if this title is suitable for your way of reading."

I think it can be removed if we have the example below.

GeorgeKerscher:
Gregorio, I think it would be best if you made the changes. You can eiter merge this one and create another PR or feel free to edit my PR. Either is fine with me.

gregoriopellegrino:
Sure, I'll wait for Gauthier's feedback and proceed.

chrisONIX:
In this paragraph:
In some jurisdictions publishers may have an exemption from the provision of accessible publications, including the provision of accessibility metadata. One example of this is from the EAA when the publisher is a micro-enterprise (i.e., enterprises employing fewer than 10 people and with annual turnover or balance sheet total not exceeding 2 million euro).

We maybe should add a few words in so it is clear that actual legal situations should be checked. 
In some jurisdictions publishers may be able to claim an exemption from the provision of accessible publications, including the provision of accessibility metadata. This should always be subject to clarification by legal counsel for each jurisdiction.
One example of this, at time of writing, is from the EAA when the publisher is a micro-enterprise (i.e., enterprises employing fewer than 10 people and with annual turnover or balance sheet total not exceeding 2 million euro).

chrisONIX:
In this paragraph:
Other legal considerations may surround exceptions, where jurisdiction will not require the title to be accessible if it involves a fundamental alteration of the content, or if making it accessible would place a disproportionate burden on the publisher.

We may want to add a few words:

Other legal considerations currently included in the EAA may include exceptions on individual versions of a title, where jurisdiction will not require the title to be accessible if it involves a fundamental alteration of the content, or if making it accessible would place a disproportionate burden on the publisher. (This may vary in other jurisdictions).

chrisONIX:
For this text:
Nevertheless, the objective is to provide as much accessibility information as is possible. It may be advisable for implementors to instead provide guidance to the end user by including a user-friendly statement such as, "This title may not yet be fully accessible to all readers. Check the list of accessibility features to determine if this title is suitable for your way of reading."

To something like this:

Nevertheless, the objective is to provide as much accessibility information as is possible. If any text relating to legal considerations needs to be displayed, then a simple phrase such as “This product is exempt from the EAA” could be used.
However it may be advisable for implementors to instead provide guidance to the end user by including a user-friendly statement such as, "This title may not yet be fully accessible to all readers. Check the list of accessibility features to determine if this title is suitable for your way of reading." 

Although these texts could be suggested int he examples

gautierchomel:
I made the changes that I found consensual. 
I suggest that we merge this one now and keep editing in new PR.

gregoriopellegrino:
That's fine with me. I would suggest writing "European Accessibility Act" in full, instead of "EAA", otherwise ok to do the merge.

gregoriopellegrino (approved):

PR #352, opened 2024-07-18
First proposal for Legal considerations in ONIX technique

gregoriopellegrino:
Based on what was defined in the principles, I created a section for Legal considerations with a placeholder where we can then insert the sentence on which we will find consensus.


gregoriopellegrino:
cc @chrisONIX

chrisONIX:
This looks good and the place holder makes sense while waiting agreement

GeorgeKerscher:
Agreed, looks good while waiting for feedback.

clapierre (approved):

gautierchomel:
In my understanding we agreed to display only one general sentence for any exemption, so there should be only one check and one sentence "TBD".

gautierchomel:
Now seeing the _update_ commit that regroups all checks into one sentence. 
Not sure about the use of discriminated checks but looks like the final is conforms to the consensus.

gautierchomel (approved):

PR #356, opened 2024-07-26
added new sections (Conformance, Legal considerations, and Additional Accessibility Information) based on ONIX Techniques

clapierre:


GeorgeKerscher:
My head hurts reading this section! Really good job, and I have one question:

Where it says: 4. ELSE:THEN display "The publication does not include a conformance statement"[ID: conformance-no].

It is checking for EPUB 1.0 and 1.1, and a, aa, and aaa. However, I believe that other conformance claims can be made. For example an audio book that meets the DAISY 3 standard.

I will review the principles again and the specification about these things.

clapierre:
@GeorgeKerscher and all  just added the new Adaptation & Clarity sections to the Additional Accessibility Information section.

clapierre:
New Legal considerations included now, all 3 new sections ready to be reviewed.

GeorgeKerscher:
I think we have a problem with the conformance section and its logic.

This is around a conformance that is not to the EPUB specification and WCAG.

If there is an accepted conformsTo  metadata, all is good.

However, we do not cover in the logic the case where there is a conformsTo to some other specification. I know publishers are including conformance statements to their own internal specifications. This is fine as long as it also conformsTo EPUB specifications. 

So, we need to cover a conformsTo statement to another specification. In the principles we have:
This publication's accessibility conformance has not been associated with accepted standards.
and we have:
The publication does not include a conformance statement.

I think the two cases above would map to:
Conformance to accepted standards for accessibility of this publication cannot be determined.

Which means all we need to add to the logic is the case where there is some text in the conformsTo metadata, but it does not match the other strings. We can then determine that there is a conformance statement, but it is not to accepted standards.

Please check my logic.

clapierre:
I like this idea George and it makes sense. 
I will take a crack at updating the EPUB technique, not sure if this will be possible in ONIX or not, as i can just look for anything in the conformsTo metadata.

I am also wondering if we should check for any other standard we know of say for audio?

clapierre:
@GeorgeKerscher I have made this improvement.

UX-Guide-Metadata/draft/techniques/epub-metadata/index.html
GeorgeKerscher, line 1: In the legal considerations  section, there is one error.
In the part with fundamental alteration, you have the text of Disproportionate burden.  That just needs to be switched out.

GeorgeKerscher:
There is one small typo in the additional information section. Where it says:
This means that a full signing of audio content of the product supplied with the video file.

Missing the word "is"

UX-Guide-Metadata/draft/techniques/epub-metadata/index.html
GeorgeKerscher, line 1: The latest changes look great.

While reviewing the clarity section, I noticed that the page break markers is checking the same OPF metadata as in the navigation section. If this is correct, I think the page navigation item in the additional features could be removed. I think they are the same, but just checking.

clapierre:
Here is a [link to the fully rendered PR](https://raw.githack.com/w3c/publ-a11y/updated-epub-metadata-sections/UX-Guide-Metadata/draft/techniques/epub-metadata/index.html).
@gregoriopellegrino take a look at the modifications to the Navigation section concerning PageNavigation and PageBreakMarkers.  I am not 100% sure if the definition is correct for PageBreakMarkers:

i.e.
page_break_markers
If true it indicates that the accessibilityFeature="pageBreakMarkers" (Print-equivalent page numbering) is present in the OPF file, otherwise if false it means that the metadata is not present.

This means that in the publication contains references to the page numbering of an equivalent printed product.

GeorgeKerscher:
This is minor, but where it says:
This means that in the publication contains references to the page numbering of an equivalent 
I would suggest:
This means that the publicationcontains references to the page numbering of an equivalent

PR #360, opened 2024-08-07
Added another example to conformance

GeorgeKerscher:
The example of Conformance to accepted standards for accessibility of this publication cannot be determined  and unknown conformance were confused.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -403,8 +403,8 @@
 						href="#conf-statements">simple statements</a> should be provided when EPUB accessibility and
 					WCAG levels are identified.</p>
 
-				<p>If the publication does not include a conformance claim,
-					the statement should indicate that conformance could not be determined. </p>
+				<p>When there is a conformance claim to an unknown specification, the statement should indicate that Conformance to accepted standards for accessibility of this publication cannot be determined. If the publication does not include a conformance claim,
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -403,8 +403,8 @@ <h3>Conformance</h3>
 						href="#conf-statements">simple statements</a> should be provided when EPUB accessibility and
 					WCAG levels are identified.</p>
 
-				<p>If the publication does not include a conformance claim,
-					the statement should indicate that conformance could not be determined. </p>
+				<p>When there is a conformance claim to an unknown specification, the statement should indicate that Conformance to accepted standards for accessibility of this publication cannot be determined. If the publication does not include a conformance claim,
+					the statement should indicate that The publication does not include a conformance statement.</p>
```
clapierre, line 407: T in The should not be capitalized.

clapierre (approved):

GeorgeKerscher:
There were three examples and I added one more. The case where there is a conformsTo to a specification other than one that we have in our list is a possibility.

I know that some publishers conform to their own internal specification. If this is addition to one we have specified, there is no problem. However, if that is the only conformsTO statement, it was not covered in the principles.

It is in the logic for the EPUB and Schema techniques. The logic is missing in the ONIX techniques.

I checked that the language strings in the principles are identical to the techniques.

PR #365, opened 2024-08-10
Updated principles before publishing

GeorgeKerscher:
I made most of the changes Matt suggested.

There is still a bit to do with this.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -114,10 +114,8 @@
 		<section id="sotd"></section>
 <section id="intro">
    <h2>Introduction</h2>
-   
-   <section id="background" class="informative">
-      <h3>Background</h3>
-      			<p>Reading a publication is a very personal experience. For most   people this is routine, and   little
+ 
+	     			<p>Reading a publication is a very personal experience. For most   people this is routine, and   little
 				consideration is given to how   the title was obtained before it is read. Users may go to a bookstore, search for the
 				title to purchase online, or have the title selected for them by an instructor for a class.</p>
 
@@ -132,7 +130,7 @@
 				accessible from the outset, not fixed later) and getting the accessibility validation or audit done by
 				independent organizations.</p>
 
-   </section>
+   
 </section>
 		<section id="general-overview">
 			<h2>General overview</h2>
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -313,8 +313,8 @@
 				</section>
 
 				<section id="va-techniques">
-					<h4>Metadata techniques</h4>
-
+					<h4>Visual adjustments metadata techniques</h4>
+<p>Specific techniques for meeting this principle are defined in the following documents:</p>
 					<ul>
 						<li><a href="../techniques/epub-metadata/index.html">EPUB Accessibility: Visual
 							Adjustments</a></li>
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I made another commit to this PR. I moved the link to the issue to just after the heading in the implementation section. 

This is the final recommendation from Matt.

There is still the question Matt poses:
The emphasis on “key” information is a bit confusing when there’s no explanation of what non-key metadata is. It might help to have an introduction to section 4 to explain why you’re only calling this metadata out as key. (This can probably wait until after you publish the first draft, though.) 

I'll make a note to myself to work on this, but a few words on what we might say would be helpful.

mattgarrish:
One thing I noticed after the review I did is you have a section called "Additional accessibility information" as a subsection of "Key accessibility information." Isn't this the non-key metadata, and should it be a sibling section and not a subsection?

Perhaps section 3 should be called "Accessibility information" and beneath that you could have "Key accessibility information" and "Additional accessibility information".

The introduction doesn't need to be long, but there must be a reason why the key metadata was chosen. You just need to say a little bit about what makes the metadata essential. I assume it's because users most often seek out this information over all the other things you could learn about a publication. Maybe you can cite that end user research went into getting to this list, or something to that effect.

Section 3.1 does a good job of explaining why the general metadata has been called out, but there isn't anything like that for the accessibility metadata.

mattgarrish:
A side note, but maybe consider removing the parenthetical "(supplement the bibliographic information)" from the section 3 heading. What that means should be explained in the section prose, although I'm kind of confused by it.

If this really is metadata that supplements the bibliographic information, then why is it calling out the publishing house, language, and publication date, which are all bibliographic information?

Also, the justification for the publication date in section 3.1 seems to contradict the point of specifying the file format. Why not have publishers state if the file is an epub 2 or 3, since that's easily determined? I doubt many users will know how dates correspond to possible formats.

GeorgeKerscher:
I removed the parenthetical  expression in the heading as matt suggested. I also added information to the file format of EPUB 2 or 3. In the description, I removed the date. I think most people will be looking at the copyright date to determine how new this title is. I also added that EPUB 2 is deprecated and that EPUB 3 supports page navigation and has better structural semantics..

GeorgeKerscher:
Based on Matt's review, I wonder if the key information section should be restructured. We have the first three listed as always being needed to be displayed. Then there is the additional information section.

Perhaps it should be:

## Accessibility information
### Key Accessibility Information
### Useful Accessibility information
### Miscellaneous  information

The miscellaneous  is the new name for additional information.

We would need a paragraph after the main heading stating something like;:

End users and educators looking for accessible titles have indicated that there are three groups of key information. There is also useful information, and miscellaneous  information that can be helpful.

mattgarrish:
I wonder if the headings would be a little easier to read if they followed a pattern like "Display techniques for ...". Calling them "metadata techniques" is a bit misleading, as this guide isn't about how to author the metadata.

For example, instead of "Supports nonvisual reading metadata techniques", have the title be "Display techniques for nonvisual reading support". (I modified the name a bit here, too, as "support" is easier to read at the end, I think.)

mattgarrish:
One more issue I just spotted: in the conformance section techniques section, the links don't have fragment identifiers at the end so they just take you to the top of their respective documents.

clapierre (approved):
Looks awesome George, I like the improvements.

GeorgeKerscher:
I am going to merge this because there are many commits. I will start a new PR to address the other comments from Matt and confirming the strings in the jason file.

PR #375, opened 2024-08-16
Implementing new function for array joining

gregoriopellegrino:
Please @clapierre check if it works for you.


UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -1062,9 +1076,8 @@ <h5>Instructions</h5>
 							<ol class="condition">
 								<li><b>LET</b> <var>adaptation_string</var> be the result of:
 									<ul class="condition">
-										<li>calling join on <var>adaptation</var> with the separator <code>", "</code></li>
-										<li>making uppercase the first character</li>
-										<li>replacing the last occurrence of <code>", "</code> with <code id="join-array-and">", and "</code>.</li>
+										<li>calling <a href="#join-array-to-comma-list">join array to comma list</a> given <var>adaptation</var></li>
+										<li>making uppercase the first character.</li>
```
clapierre, line 1080: Where does that uppercase of the first character happen?
I don't see that in the function itself nor its example of
`<p>For example if the input array has elements <code>["one", "two", "three"]</code>, then the output is going to be <code>"one, two, and three"</code>.</p>`

mattgarrish: This seems okay to me. This step uppercases the string returned from the function.

You might just want to say "uppercasing the first character" for simplicity, though.

gregoriopellegrino: Done, thank you

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -825,9 +840,8 @@ <h4>Instructions</h4>
 							</li>
 							<li><b>LET</b> <var>hazards_string</var> be the result of:
 								<ul class="condition">
-									<li>calling join on <var>hazards</var> with the separator <code>", "</code></li>
-									<li>making uppercase the first character</li>
-									<li>replacing the last occurrence of <code>", "</code> with <code id="join-array-and">", and "</code></li>
+									<li>calling <a href="#join-array-to-comma-list">join array to comma list</a> given <var>hazards</var></li>
+									<li>uppercasing the first character</li>
```
clapierre, line 844: Here too talks about uppercasing the first character.

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -1062,9 +1076,8 @@ <h5>Instructions</h5>
 							<ol class="condition">
 								<li><b>LET</b> <var>adaptation_string</var> be the result of:
 									<ul class="condition">
-										<li>calling join on <var>adaptation</var> with the separator <code>", "</code></li>
-										<li>making uppercase the first character</li>
-										<li>replacing the last occurrence of <code>", "</code> with <code id="join-array-and">", and "</code>.</li>
+										<li>calling <a href="#join-array-to-comma-list">join array to comma list</a> given <var>adaptation</var></li>
+										<li>uppercasing the first character.</li>
```
clapierre, line 1080: Sam here about uppercase

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
```diff
@@ -1162,9 +1175,8 @@ <h5>Instructions</h5>
 							<ol class="condition">
 								<li><b>LET</b> <var>clarity_string</var> be the result of:
 									<ul class="condition">
-										<li>calling join on <var>clarity</var> with the separator <code>", "</code></li>
-										<li>making uppercase the first character</li>
-										<li>replacing the last occurrence of <code>", "</code> with <code id="join-array-and">", and "</code>.</li>
+										<li>calling <a href="#join-array-to-comma-list">join array to comma list</a> given <var>clarity</var></li>
+										<li>uppercasing the first character.</li>
```
clapierre, line 1179: Here too.

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
gregoriopellegrino: That's right, it is a sub-level list, where in the variable assignment the array is merged and the first letter is capitalized, this is a mode we already had before the introduction of the new function and was suggested by @mattgarrish

For example:
<ol><li><b>LET</b> <var>hazards_string</var> be the result of:
								<ul class="condition">
									<li>calling <a href="#join-array-to-comma-list">join array to comma list</a> given <var>hazards</var></li>
									<li>uppercasing the first character</li>
									<li>concatenating <code id="hazards-plural">" hazards"</code> to the end of the string <b>IF</b> (length of <var>hazards</var>) > 1 <b>ELSE</b> concatenating <code id="hazards-singular">" hazard"</code> to the end of the string.</li>
								</ul>
							</li></ol>

gregoriopellegrino: I cannot insert uppercasing within the `join array to comma list` function, because the generated string does not always have to be capitalized at the beginning.

mattgarrish: Could you explain what your concern is, Charles? Uppercase functions are pretty ubiquitous, and almost never manually coded, so I don't expect there's any value in documenting how that works.

There's probably an argument that uppercasing isn't applicable to every language, but I don't know that that needs stating. I'm sure anyone implementing a display transform for such languages will already understand that.

clapierre: I don't care if we capitalize or not, my concern is that in the code it says this function will capitalize the first word which it doesn't.

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
clapierre: Like here it says it is going to capitalize the first letter, but if it doesn't we are confusing people so we should just remove that line.

gregoriopellegrino:
The way I thought of it, at @mattgarrish's suggestion, `uppercasing the first character` is a generic function that is called after `join array to comma list` returns the string.

If I were to implement it in JavaScript it would be like:
```javascript
hazards_string = joinArrayToCommaList(hazards).capitalize() + " hazards";
```

UX-Guide-Metadata/draft/techniques/onix-metadata/index.html
mattgarrish: It doesn't say that the function will capitalize. It's a separate list item, so a separate instruction to whoever is implementing this step to uppercase the first letter.

Would it be clearer for you if it said "Uppercasing the first character in the resulting string"?

clapierre:
Thats fine.

PR #377, opened 2024-08-16
Repo cleanup

gregoriopellegrino:
I did some cleanup in the draft folder:
- removed images used in version 1.0 and no longer used in version 2.0
- deleted the original files (HTML exported from Google Doc) from which we had started the work of version 1.0
- updated the references in the readme


clapierre (approved):

UX-Guide-Metadata/draft/README.md
```diff
@@ -1,6 +1,6 @@
-# User Experience Guide for Displaying Accessibility Metadata
+# User Experience Guide for Displaying Accessibility Metadata 2.0
  
 * [Current Working Draft - Principles](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/draft/principles/)
```
mattgarrish, line 3: I'd suggest making sure you're not using "working draft" to describe these documents, like here, as that's a term for working group documents. You should call these editor's drafts.

When published they're either draft CG notes, if you're still planning to work on them, or CG notes if they're the final docs.

PR #378, opened 2024-08-17
Changes to Principles from August 16 call

GeorgeKerscher:
This improves the language in navigation supported by the page list. Also softened language around conformance to an unknown specification. I did not mention an optimized specification because I did not think most people would know what that means. Added the note before the examples in additional information that there many features and we should some. Complete list in techniques. I removed sign language which were associated with closed captions. Close captions will be common and sing language videos will be rare.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -893,14 +893,15 @@ Always get legal advice from in-house or your retained counsel to assess legal c
 
 				<section id="add-info-examples">
 					<h2>Examples</h2>
-
+<div class="note"	>There are many features that may be present.  Some of the features are listed in the examples. A complete list can be found in the techniques. 
+</div>
 					<aside class="example" title="Descriptive explanations">
 						<ul>
 							<li>Content is enhanced with ARIA roles to optimize organization and facilitate
 								navigation.</li>
 <li>Page breaks   included  from the original print source.</li>
-							<li>Videos included in publications have closed captions and sign language representation. </li>
-							<li>Tactile graphics have been integrated to facilitate access to visual elements by blind
+							<li>Videos included in publications have closed captions. </li>
+							<li>Tactile graphics have been integrated to facilitate access to visual elements for blind
 								people.</li>
 							<li>Music scores are provided that are visually accessible.</li>
 							<li>Content contains information encoded through colors, which may be imperceptible to those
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -398,7 +398,7 @@ <h3>Conformance</h3>
 						href="#conf-statements">simple statements</a> should be provided when EPUB accessibility and
 					WCAG levels are identified.</p>
 
-				<p>When there is a conformance claim to an unknown specification, the statement should indicate that Conformance to accepted standards for accessibility of this publication cannot be determined. If the publication does not include a conformance claim,
+				<p>When there is a conformance claim to an unknown specification, the statement should indicate that a    Conformance statement is present and it is to    an unknown specification. If the publication does not include a conformance claim,
```
mattgarrish, line 401: This comment isn't specifically about this change, but I really wonder about the usefulness of reporting unknown standards while providing no detail about them. The conformsTo metadata can reference any standard, not just an accessibility one, so what help is it to tell users the publication conforms to an unknown standard? I doubt presenting random strings in conformsTo is a good idea, so I'd argue in this case it's better to say nothing at all.

It might make more sense to say that vendors can decide for themselves whether to present conformance information they recognize but that isn't detailed in these guides, but otherwise the onus should probably be on people conforming to unrecognized standards to bring them to our attention to vet them for display.

GeorgeKerscher:
We know that ONIX does not have a code for an unknown specification and it was considered and they decided not to have a code at this time. 

Removing this and the example would resolve the issue for ONIX.

We have PDF/UA in ONIX code 05. I don't know that we have similar  metadata in EPUB accessibility I think we thought our guidelines would apply to PDF as well. 

If we have a value in schema.org for PDF/UA  or can create one in our vocabulary, should we put this into the examples of conformance?  

It is probably not the right time to add more specifications, but if libraries serving people with disabilities will use this, we should add DAISY 2.202 and 3 to the list of known specifications.

mattgarrish:
> If we have a value in schema.org for PDF/UA or can create one in our vocabulary

No, conformance claims aren't part of the schema.org metadata. I think it's outside the scope of our work here, too, as it would necessitate a PDF techniques document, as there'd have to be a way to read the conformance from the PDF file.

clapierre (approved):
I think I agree with Matt, that maybe we should not say anything about conformsTo declarations we do not know, and include a note to encourage folks to file a GitHub issue if there is a standard worth including in our approved list.

avneeshsingh:
we can remove example of conformance to unknown standard. We can revisit it when we have something concrete in ONIX.

clapierre (approved):

PR #382, opened 2024-08-29
Proposed folder structure for first public draft release

gautierchomel:


gregoriopellegrino (approved):

mattgarrish:
Sorry for the late comment, but the "public-draft" folder only appears to add another unnecessary layer of structure. All that's going to do is lead to more duplication when you have to publish the final versions in yet another set of folders.

When we publish drafts on the path to a final version, the typical way to do it is to have two files under the final directory - the current version goes in index.html and a dated version goes in index-yyyymmdd.html.

When you publish the next draft note, that new version replaces the one in index.html but the dated draft you created doesn't get overwritten as you create a new dated file. In the end, you'll have an index.html with the final note and all the dated versions that proceeded it together in one directory

In other words, you only need /UX-Metadata-Guide/2.0/principles/ for all the drafts and final note for the principles, with similar for the techniques.

Issue #371, opened 2024-08-13
Charts Diagrams and Formulas examples are not aligned between Principles and Technics.


gautierchomel:
Principles has only two examples for Charts, Diagrams and Formulas : 

1. Accessible formulas. Charts and diagrams have extended descriptions.
2. Accessibility of formulas, charts, and diagrams unknown.

While technics have 5 : 

1. "Charts and diagrams have extended descriptions".
2.   "Visualized data also available as non-graphical data".
3.  "Accessible chemistry content".
4.  "Accessible math content".
5. "Accessibility of formulas, charts, and diagrams unknown".

gautierchomel:
I aligned the examples in principles to fit.

Issue #369, opened 2024-08-13
ONIX tecnics refers to Unknown Audio but this is not present in Principles


gautierchomel:
ONIX technics refers to [No information about pre-recorded audio](https://github.com/w3c/publ-a11y/blob/e50ad271f3e41b391d4c62d62b1820bfe4a830e2/UX-Guide-Metadata/draft/techniques/onix-metadata/index.html#L574) but this example does not exist in the principles.

gautierchomel:
Not sure I understand well, but as it is I feel that millions of textual ebooks would have this mention, which would be confusing.

GeorgeKerscher:
Correct, we do not want to say that no audio is present. I believe this has been corrected in the latest version of the guide.

Issue #368, opened 2024-08-13
Conformance but not EPUB exist in principles but not in Technics


gautierchomel:
In principles there is a sentence  [The publication contains a conformance statement that is not to the EPUB Accessibility](https://github.com/w3c/publ-a11y/blob/e50ad271f3e41b391d4c62d62b1820bfe4a830e2/UX-Guide-Metadata/draft/principles/index.html#L432) That is not documented in the Technics.

Issue #367, opened 2024-08-13
Non Visual Reading Unknown exist in principles but not in technic


gautierchomel:
In principles there is a sentence [Not known if readable in read aloud and braille](https://github.com/w3c/publ-a11y/blob/e50ad271f3e41b391d4c62d62b1820bfe4a830e2/UX-Guide-Metadata/draft/principles/index.html#L371) That is not documented in the Technics.

GeorgeKerscher:
Commented out the fourth string in principles. The commit will close this issue once merged.

PR #384, opened 2024-08-29
Ready to merge Localization updates

gautierchomel:
This includes: 

- updated Keys in the canonical.json file for localization
- additional IDs to principles examples so they are identifiable horizontally (same ID in Principle, Technics and JSON)

Raised some issues on the walkway: 

- #392 
- #391 
- #390 
- #371 (solved by aligning Principles to hasard keys)
- #369 
- #368 
- #367


gautierchomel:
I suggest we merge so the principles get the IDs and make any further improvements to the JSON later.

clapierre (approved):
Nice work.

gautierchomel:
I implemented one upper level to group keys. 

"detailed-Conformance" remains the only object that has an additional layer, that's due to the structure of the Conformance section.

GeorgeKerscher:
In the principles document on lines 500 and 527, the span should be closed after the</a>.

gautierchomel:
> In the principles document on lines 500 and 527, the span should be closed after the.

Done for those two and three more occurrences. 

That means the "." must be in the JSON value, am I right?

PR #385, opened 2024-08-30
One less folder level as suggested by Matt

gautierchomel:
> When you publish the next draft note, that new version replaces the one in index.html but the dated draft you created doesn't get overwritten as you create a new dated file. In the end, you'll have an index.html with the final note and all the dated versions that proceeded it together in one directory


gregoriopellegrino (approved):

mattgarrish (approved):
LGTM

clapierre (approved):

PR #387, opened 2024-08-30
Fixing broken links for first public draft

gregoriopellegrino:


clapierre (approved):
Thanks, Gregorio, for also fixing the EPUB techniques.

PR #393, opened 2024-08-31
Create README.md

gautierchomel:
An initial proposal to explain the Localization process and invite more local projects for feedback collection.


gregoriopellegrino:
I like the introductory part, but I don't know if it is suitable for this type of document, it seems to me more for the Principles document, for example in the implementations section.

Some notes:
- We had assumed to use an external tool to contribute, right? Or will people have to write the JSON?
     - In the second case maybe better to explain with an example what we mean by changing the value, leaving the key unchanged
- For any questions I would leave the link to the issues
- I think it would make sense to put the links to the principles and techniques

gregoriopellegrino:
> > We had assumed to use an external tool to contribute, right? Or will people have to write the JSON?
> 
> Both ways are possible.

Okay, perfect. Then maybe put the link to Gitlocalize?

clapierre (approved):

gautierchomel:
> maybe put the link to Gitlocalize?

Done.

PR #409, opened 2024-09-20
Note regarding the use of "compact" vs. "descriptive"

clapierre:


GeorgeKerscher:
Should you explain that the ID in the JSON file takes you to a group that has both the compact and the descriptive, and it  is here that you are free to choose?

clapierre:
I can @GeorgeKerscher , would this come at the bottom of the note, and what exactly should it say?

Currently after the example I have:

 The option is yours to provide either the "compact" or "descriptive" versions to the end user or perhaps even allow the user to choose what they wish to see.

GeorgeKerscher:
I think the only thing you need to add is that the ID linking mechanism  thakes you to both the  compact and descriptive, and then you choose, i.e. the ID is the same.

gregoriopellegrino:
For me a good starting point. Perhaps in the example we could use an excerpt from the JSON instead of the principles?

We could also mention that there is JSON and to refer to the Localization section in the Principles document.

gautierchomel (approved):

clapierre:
Hello all, Ok, I have updated the Example to use the real JSON, and added the editors note as requested by @mattgarrish.
If this looks good We can merge this and @gregoriopellegrino can copy this into the ONIX techniques.

gregoriopellegrino:
In this same branch I added the note to the ONIX techniques.

gregoriopellegrino (approved):

PR #416, opened 2024-10-09
Changed example 2 from WCAG 2.0 to 2.2

GeorgeKerscher:
I need confirmation on what I am doing.

In the conformance section, I changed example 2 which had WCAG 2.0 to 2.2. I also changed the id. In the excel spreadsheet I saw that conformance-wcag-2-2 had 0 and 2-0 has one.

I did not make any modification to the spreadsheet. I did not add the spreadsheet to the commit either.

Should I update the spreadsheet somehow? I thought this was being auto generated.

Have I done this correctly?

In the next phase where I am adding information, I also need to add the correct ID as well, right?


clapierre (approved):

gregoriopellegrino:
Hi @GeorgeKerscher it seems to me that you have done a very good job. After the merge, I will take care of updating the Excel file using the script.

PR #431, opened 2024-10-15
Updated readme

mattgarrish:
This beefs up the readme for the repository by focusing it on the task force and the documents we produce rather than as a generic repository for discussions in the publishing activity. I added a section on participating in the task force and listed all the documents we host here.

This will also gives us a better reference point for the rechartering, as we may move things like pageBreakSource into the spec in a new revision.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,11 +1,44 @@
 
 ![W3C Logo](https://www.w3.org/Icons/w3c_home)
 
-# Accessibility Discussions of the Publishing@W3C Groups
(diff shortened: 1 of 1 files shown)


mattgarrish:
I can't get a full html preview, but here's the raw markdown: https://raw.githack.com/w3c/publ-a11y/update-readme/README.md

gregoriopellegrino (approved):

clapierre (approved):
nice work Matt.

PR #433, opened 2024-10-16
Added e-book in several places and other minor changes

GeorgeKerscher:
It seems that the EAA uses e-book and e-readers in the language of the directive. I introduced these terms and use e-book and digital publication interchangeably. I have not defined these terms or made a statement that they are used interchangeably.

If this is needed, let me know.

I also looked for areas where we could make it less EPUB centric. I made some small changes. It seems thatPDF, or EPUB are neutral, until you come to the conformance section. We do not have an example of a PDF. I would need some help in writing such an example.

Perhaps a PDF conformance statement should be added after we have a techniques section to support internal PDF metadata. Alternatively,it might be possible to craft a conformance statement for PDF from the ONIX metadata.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -117,20 +117,20 @@
 		<section id="intro">
 			<h2>Introduction</h2>
 
-			<p>Reading a publication is a very personal experience. For most people this is routine, and little
+			<p>Reading a digital publication is a very personal experience. For most people this is routine, and little
 				consideration is given to how the title was obtained before it is read. Users may go to a bookstore,
 				search for the title to purchase online, or have the title selected for them by an instructor for a
 				class.</p>
 
 			<p>Now consider that the person is blind and relies on assistive technology. The user needs that technology
-				to assist them in the purchase process as well as to read the publication. The person may wonder: will
+				to assist them in the purchase process as well as to read the e-book. The person may wonder: will
 				the screen reader work with this title; are there image descriptions that will be spoken to describe
 				these images; are there page numbers which are accessible; is the reading order correct so a caution
 				after reading a paragraph which could be dangerous will be announced? All of these, and more
 				accessibility concerns are potential issues consumers have when trying to purchase and ultimately read a
-				digital publication.</p>
+				digital publication in any format.</p>
 
-			<p>The good news is more and more publishers are creating publications that are Born Accessible (i.e.,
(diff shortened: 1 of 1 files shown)


mattgarrish:
W3C style is to US spellings and ebook (or eBook) is more common, at least that I've ever found in publishing retail. Hyphenated "e" things make me want to party like it's 1999.

GeorgeKerscher:
Based on this comment and knowing that it is a broad not only W3c audience, I am inclinedd to add a note. It would clarify that 
digital publications are descriedd in many ways, including: e-books, eBooks, EPUB, PDF publications, audio books, DAISY books,  PDF publications, and more. It should clarify that it is a digital version that is intended to be used through a computer or a device, and that persons with disabilities use Assistive Technologies to access this information through mainstream Reading Systems or through technologies optimized for their .particular disability.

mattgarrish:
The benefit of digital publication is that it avoids the book baggage, specifically being tied to text. Since it's in the title of the document, we should be consistent in using that term unless there's a specific case where we need to call out a particular format. A definition that clarifies it's an overarching terms for all the types of publications you've listed makes sense to me.

gregoriopellegrino (approved):

clapierre (approved):
Like Matt says there are many places here where e-book is present also e-reading device.
Should this also be changed to eReading device? or just electronic reading device?

mattgarrish:
Isn't eReader the more common US name?

clapierre:
Hi Matt, I believe so. So George, I think it would be best to change e-Reading device to eReader. which in theory you could say

eReader (electronic reading device) if you wanted to.

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -130,22 +130,22 @@ <h2>Introduction</h2>
...
-			<p>The good news is more and more publishers are creating e-books that are Born Accessible (i.e.,
+			<p>The good news is more and more publishers are creating digital publications that are Born Accessible (i.e.,
 				accessible from the outset, not fixed later) and getting the accessibility validation or audit done by
 				independent organizations.</p>
 		</section>
 		<section id="general-overview">
 			<h2>General overview</h2>
 
-			<p>This document helps those who wish to render accessibility metadata directly to users understand how to
+			<p>These guidelines   helps those who wish to render accessibility metadata directly to users understand how to
```
mattgarrish, line 140:
```suggestion
			<p>These guidelines   help those who wish to render accessibility metadata directly to users understand how to
```

PR #434, opened 2024-10-17
Updated overview diagram

mattgarrish:
I'm only opening this as a draft pull request as I expect we may still need to make some adjustments. It'd be good to get feedback from the other editors before moving on to a proper pull request.

To recap, section 2.2 is renamed "Metadata process overview" and the text has been revised to describe how these documents fit in with how metadata arrives and is processed. It's not a workflow, specifically, but more an explanation of the big picture of what we're trying to solve.

* [Preview](https://raw.githack.com/w3c/publ-a11y/fix/issue-397/UX-Guide-Metadata/draft/principles/index.html)
* [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/UX-Guide-Metadata/draft/principles/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/fix/issue-397/UX-Guide-Metadata/draft/principles/index.html)


mattgarrish:
> I would like to see a better word instead of "precedence "

I've changes this to:

>   and prioritizes the display for readers

Do you think that's more accurate, George?

GeorgeKerscher:
Yes, I think that is better. It is really like a decision tree or something like that. Prioritizes  is as good as I can think of.

GeorgeKerscher:
I suggest   putting the However at the beginning of the sentence in: The process of transforming the metadata is out of scope of this document, however. I also like the description that follows the image. Not having it in details element is also good; I think everybody benefits from the description.

mattgarrish:
> I suggest putting the However at the beginning of the sentence in:

Ya, we got beaten with Strunk and White through university to never start a sentence with however. I can't kick the habit.

> Not having it in details element is also good

Right, I didn't like that the whole section depended on seeing the image or having to expand the description. I think it's better to use the image as the basis for the narrative of the section.

mattgarrish:
An open question we had revising the document is what to reference for MARC? The image currently references the fields where the accessibility is recorded, but is there a standard and/or techniques document like for EPUB, PDF and ONIX that we can cite?

(Or is it too early to be integrating MARC? It's noted as under development, but is it likely to be in the 2.0 guide before we publish?)

GeorgeKerscher:
Just spent some time with Grammar Girl and her guidance with starting sentences with However. So, it is acceptable to start a sentence with "however" but you need to be careful.

We already have a "but" in this paragraph. One option would be to just use a semicolin to join the two clauses and drop the however at the end.

Oh, and Grammar Girl is on the Read 2.0 list I follow. 

What ever you decide, I am sure it will be fine.

avneeshsingh:
I think IFLA base camp will need some time, I do not see MARC techniques to be ready in this year or in initial weeks of next year.

We can check with our key contacts.

mattgarrish:
In the end, I reworked the note to put the out of scope statements together, so dropped the "however". It now reads:

>This guide assumes the metadata is already in one of the formats described in the diagram, but depending on how the metadata is submitted it may need to be transformed (e.g., if a vendor prefers to handle only ONIX metadata for display they would need to preprocess the metadata embedded in an EPUB or PDF file).
>
> The process of transforming metadata is not it scope for this document, nor is how to reconcile metadata when it is provided in multiple forms. The guide assumes any processing has already occurred. For information on mapping between formats, refer to the <a href="https://w3c.github.io/publ-a11y/drafts/a11y-crosswalk-MARC/">Accessibility Properties Crosswalk</a>.

mattgarrish:
> I do not see MARC techniques to be ready in this year or in initial weeks of next year

It might make sense to go back to an earlier graphic when we didn't mention MARC in that case.

GeorgeKerscher:
Christine Oliver <[Christine.Oliver@uottawa.ca](mailto:Christine.Oliver@uottawa.ca)>

Responded and George is pasting it here. She writes:

Thank you for the heads-up about this question. The IFLA Accessibility Metadata Network is working on the mapping to UNIMARC and MARC 21. I am not on that working group, but I don't think they will be completing the work in 2024. There was a preliminary draft prepared for the W3C group chaired by Avneesh, but we needed a larger group to analyze and validate the preliminary work. Now we have that larger group, and they are actively working on this mapping or crosswalk.

I can check with the chair of this working group and let you know if there is a projected timeframe.  

Best wishes,

Chris

mattgarrish:
As discussed on the editor's call, I've moved the editor's note about PDF and MARC support not being finalized to a regular note after the diagram. We don't need to modify the graphic anymore this way.

PR #441, opened 2024-10-22
Full stop and more

GeorgeKerscher:
Thank you @Gregorio  the list of full stops made it easy to remove them. That was one commit.
I filed an #issue-438 about the strings and have some disagreements.
I filed an issue 440 about thefourth example in conformance is missing IDs.

I removed a lot more full stops in the last two commits.

I will file an issue that the additional information missing IDs

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -242,7 +242,7 @@
 
 				<div class="note">
 					<p>This guide assumes the metadata is already in one of the formats described in the diagram, but
-						depending on how the metadata is submitted it may need to be transformed (e.g., if a vendor
+						depending on how the metadata is submitted it may need to be transformed (e.g.	, if a vendor
 						prefers to handle only ONIX metadata for display they would need to preprocess the metadata
 						embedded in an EPUB or PDF file).</p>
 					<p>The process of transforming metadata is not it scope for this document nor is how to reconcile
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -416,26 +416,26 @@ <h4>Examples</h4>
...
 								modified according to the capabilities of the reading system (font family and font size,
 								spaces between paragraphs, sentences, words, and letters, as well as color of background
-								and text).</li>
+								and text)</li>
 							<li data-localization-id="visual-adjustments-unmodifiable"
-								data-localization-mode="descriptive">Text and page layout cannot be modified. The
+								data-localization-mode="descriptive">Text and page layout cannot be modified The
 								reading experience is close to a print version. Reading systems can still provide
-								zooming options.</li>
+								zooming options</li>
```
mattgarrish, line 423: This is highly problematic. Should the strings all be a single statement? Taking a period out of the middle of a paragraph can't be done, and presumably the other period here would have to be removed. Could you use commas here, perhaps, so make it more like:

> Text and page layout cannot be modified as the reading experience is close to a print version, but reading systems can still provide zooming options

mattgarrish (changes requested):

GeorgeKerscher:
I agree. making changes now. Will look for more of these things.

GeorgeKerscher:
In conformance we have:
Detailed Conformance Information:
This publication claims to meet EPUB Accessibility 1.1 Web Accessibility Content Guidelines (WCAG) 2.1 Level AA The publication was certified on 2021-09-07 by Foo's Accessibility Testing with a credential of "Enterprise Accessibility Rating" For more information, refer to the certifier's report

Would it work better to make these list items? 
 statement that claims to meet accepted accessibility standards followed by detailed conformance information">
						<p data-localization-id="conformance-aa" data-localization-mode="compact">This publication meets
							accepted accessibility standards</p>
						<p>
							<span data-localization-id="conformance-certification-info" data-localization-mode="compact"
								>The publication was certified</span>
							<span data-localization-id="conformance-certifier-prefix" data-localization-mode="compact"
								>by</span> Foo's Accessibility Testing </p>
						<p>

mattgarrish:
I'm kind of confused reading that section. The examples aren't split like the other with compact and extended descriptions. Each begins with three statements and then a heading "Detailed Conformance Information:" followed by a paragraph.

Is it correct that the first three statements are what you use as the compact statements, or are they examples of what information you use to write the descriptive statement that follows. Also, is it an omission that [example 8](https://w3c.github.io/publ-a11y/UX-Guide-Metadata/draft/principles/#example-unknown-conformance-statement) only has the first three statements and no descriptive statement?

And in example 7,. it says the conformance cannot be identified, but in the descriptive statement there's a placeholder for inserting a publisher's internal conformance standard. Shouldn't that be removed?

But back to your question, if the goal is to omit the periods, I think you have to break the paragraphs up into separate statements. That would let the implementor decide whether to format them as a list or combine them into a paragraph using periods. I don't see any easy way to combine them all into one without making a large run-on sentence.

gregoriopellegrino (approved):

clapierre (approved):
Once all merged, we can make sure no period was taken out between paragraphs.

PR #451, opened 2024-10-29
Made changes as discussed on October 29

GeorgeKerscher:
One descriptive example was provided. I made it singular. Where it talks about what was provided, I say that in most cases this will not be displayed, but where it is, we use the string: "This publication claims an accessibility exemption in some jurisdictions"

I will comment in issue 350 about this approach.

### DIFF
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -1102,13 +1102,10 @@
 
 				<section id="legal-examples">
 					<h4>Example</h4>
-
-					
-<p>In many cases no legal exception or exemption information will be presented. Where it it is provided, the example is  provided as   a possible descriptive   explanation.</p>
- 
-					<aside class="example" title="Descriptive explanation">
+		<aside class="example" title="In many cases no legal exception or exemption information will be presented. The following example is provided as a possible descriptive explanation when legal exception or exemption information needs to be presented.">
 
 <p data-localization-id="legal-considerations-exempt"
+
 								data-localization-mode="descriptive">This publication claims an accessibility exemption in some jurisdictions</p>
 </aside>
 
--- a/UX-Guide-Metadata/draft/principles/index.html
+++ b/UX-Guide-Metadata/draft/principles/index.html
@@ -1102,7 +1102,7 @@
 
 				<section id="legal-examples">
 					<h4>Example</h4>
-		<aside class="example" title="Exception or exception ">
+		<aside class="example" title="Exception or exemption ">
 <p> In many cases no legal exception or exemption information will be presented. The following example is provided as a possible descriptive explanation when legal exception or exemption information needs to be presented.</p>
 <p data-localization-id="legal-considerations-exempt"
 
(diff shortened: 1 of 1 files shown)


UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -1101,17 +1101,17 @@ <h3 data-localization-id="legal-considerations-title">Legal considerations</h3>
 					is to provide as much accessibility information as is possible.</p>
 
 				<section id="legal-examples">
-					<h4>Examples</h4>
-					<p>TBD</p>
-					<!--
+					<h4>Example</h4>
+
+					
+<p>In many cases no legal exception or exemption information will be presented. Where it it is provided, the example is  provided as   a possible descriptive   explanation.</p>
```
mattgarrish, line 1107: I think this could read more simply as: "The following example provides as a possible descriptive explanation when legal exception or exemption information needs to be presented."

At a minimum, there is an "it it" that needs fixing in the original.

```suggestion
<p>The following example provides as a possible descriptive explanation when legal exception or exemption information needs to be presented.</p>
```

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -1101,17 +1101,17 @@ <h3 data-localization-id="legal-considerations-title">Legal considerations</h3>
...
+ 
 					<aside class="example" title="Descriptive explanation">
```
mattgarrish, line 1109: It might be a bit strange to have a singular "Example" title for the section followed by a description of the example followed by the example box which will generate another example heading. Usually when a section is an example you don't use a box. If you want to keep this formatting, I'd put the description for the example inside the box, as well.

mattgarrish (changes requested):

mattgarrish:
The box is what gets visually constructed when respec turns the aside into the example. It's sort of like a sidebar. The example gets a yellow shaded box with a darker yellow left border.

What I meant was to move the p tag with the description inside the aside so that it will be part of that yellow box rather than before it. By adding the text to the title attribute, that text will now be merged with the auto-generated example number as the heading. It's going to be too long.

GeorgeKerscher:
Matt, let me know if these changes are what you expect. I don't know what a box is. Is that what an aside looks like?

gregoriopellegrino (approved):

UX-Guide-Metadata/draft/principles/index.html
```diff
@@ -1101,17 +1101,14 @@ <h3 data-localization-id="legal-considerations-title">Legal considerations</h3>
 					is to provide as much accessibility information as is possible.</p>
 
 				<section id="legal-examples">
-					<h4>Examples</h4>
-					<p>TBD</p>
-					<!--
-					<aside class="example" title="Descriptive explanation">
+					<h4>Example</h4>
+		<aside class="example" title="Exception or exception ">
```
mattgarrish, line 1105:
```suggestion
		<aside class="example" title="Exception or exemption ">
```

mattgarrish:
Yes, it looks good now. I just made a minor fix to the title as it was "Exception or exception". I expect that should have been "Exception or exemption".

gregoriopellegrino (approved):

mattgarrish (approved):

PR #453, opened 2024-10-30
Made changes in conformance that resolves presentation

GeorgeKerscher:
I simply added a few paragraphs that resolves how the conformance reads. Simple fix, but hard to figure out.

While doing this, it triggered some questions I will post an issue for.


gregoriopellegrino:
No string changes, so we do not need to update techniques

gregoriopellegrino (approved):

clapierre (approved):

PR #455, opened 2024-10-30
Made changes in the language around refreshable and electronic braille

GeorgeKerscher:
I went through the document and clarified the terms. It now says generated refreshable braille in many places.

In the compact, I just left it as braille.

I will now go into the techniques to match up the strings there. I will commit them to this PR.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -167,7 +167,7 @@
 <dt><dfn>electronic braille</dfn></dt>
 <dt><dfn>refreshable braille</dfn></dt>
 <dd>
-<p>The terms "electronic braille" and "refreshable braille" are used interchangeably, which denotes a device with pop-up pins to present the braille on a tactile screen. These devices can be used as a display for a personal computer, or they may be a self-contained multipurpose note taker. </p>
+<p>The terms "electronic braille" and "refreshable braille" can be  used interchangeably, which denotes a device with pop-up pins to present the braille on a tactile screen. These devices can be used as a display attached to  a personal computer, or they may be a self-contained multipurpose note taker with a refreshable braille display. </p>
 </dd>
    </dl>
 </section>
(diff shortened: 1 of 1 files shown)


a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -167,7 +167,7 @@ <h3>Terminology</h3>
 <dt><dfn>electronic braille</dfn></dt>
```
mattgarrish, line 167: If we don't use a term in a document, we can't define it in the terminology. If you want to compare the terms in the definition, that's fine, but you need to remove the one that isn't being used in the documents. That would be electronic braille now, from what I can tell.

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -167,7 +167,7 @@ <h3>Terminology</h3>
...
 <dt><dfn>refreshable braille</dfn></dt>
 <dd>
-<p>The terms "electronic braille" and "refreshable braille" are used interchangeably, which denotes a device with pop-up pins to present the braille on a tactile screen. These devices can be used as a display for a personal computer, or they may be a self-contained multipurpose note taker. </p>
+<p>The terms "electronic braille" and "refreshable braille" can be  used interchangeably, which denotes a device with pop-up pins to present the braille on a tactile screen. These devices can be used as a display attached to  a personal computer, or they may be a self-contained multipurpose note taker with a refreshable braille display. </p>
```
mattgarrish, line 170: This still sounds like a definition of a refreshable braille display. That's different from the generated braille itself, which is what the term is used for in the body. To avoid confusion, we might want to use a term like "dynamic braille" for what gets rendered.

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -489,16 +482,14 @@ <h4>Examples</h4>
 					<aside class="example" title="Descriptive explanations">
 						<ul>
 							<li data-localization-id="nonvisual-reading-readable" data-localization-mode="descriptive"
-								>All content is available in generated read aloud speech and electronic braille</li>
+								>All content is available in generated read aloud speech and generated refreshable  braille</li>
```
mattgarrish, line 485: This still sounds like it's implying that the speech and braille have been pre-generated and are available to everyone. These options only exist if the device has those capabilities, and you may only be able to get one or the other. It would make more sense to me to say "All content can be generated as read aloud speech or dynamic braille"

Similarly, for the next two I'd go with

- Portions of the content may not be generable as read aloud speech or dynamic braille
- Not all of the content can be generated as read aloud speech or dynamic braille

mattgarrish (changes requested):
It might be better to leave this until after the coming release so we can discuss it some more.

gregoriopellegrino:
This change only impacts the descriptive strings, so we don't have to make changes in the techniques. OK for me.

gregoriopellegrino (approved):

GeorgeKerscher:
I agree that •	Not all of the content can be generated as read aloud speech or dynamic braille is a good improvement.

We would need to add dynamic  in the definations and in various places.

Should we merge and refine the language after the next release?

The compact strings would not change and when we make any agreed upon changes, it should be easy to update. The descriptive strings would change in the JSON file.

clapierre (approved):

PR #456, opened 2024-11-01
Remove redirects to new directory

mattgarrish:
Merging this pull request will delete the redirects on the editor drafts and draft notes that were published under the old UX-Guide-Metadata directory.

Opening this so we don't forget to remove them, but don't merge this PR until after our next release of revised draft notes (i.e., not the ones we're about to release). No one should still be going to these old file locations by that point.


GeorgeKerscher:
the new title will be "Accessibility Metadata Display Guide for Digital Publications " correct? I guess the old name is being deleted by this pr.

mattgarrish:
This only removes the redirects so anyone who bookmarked the 2.0 documents in the UX-Guide-Metadata directory ends up at the updated documents.

The changes we've already made are reflected in the documents in the /a11y-meta-display-guide directory, so the new title is in place for all future releases.

mattgarrish:
FYI, the first drafts we published back in September have the old name because that's what we were using then. That's why you will still see the old title if you go to the current published documents, even though they're under the new directory structure. That will change once we publish the new drafts today (or whenever they go out).

mattgarrish:
I'm going to go ahead and take out these old redirects now. It's been almost three months, so that's plenty of time to have updated bookmarks.

Issue #475, opened 2024-11-13
Improve consistency in metadata display strings


gregoriopellegrino:
While gathering feedback on the Accessibility Metadata Display Guide 2.0, we noticed some inconsistencies in how features are described. For example, in the Navigation section (https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/guidelines/#nav-examples), some strings begin with "Contains..." while others directly describe the feature (e.g. "Index provides links...").

We think it would be valuable to establish a consistent pattern for all descriptive strings across the guide. This would make the information more predictable and easier to scan, while also helping with localization efforts.

GeorgeKerscher:
I think we have addressed this issue with the various PR through 639.

PR #492, opened 2024-11-22
Update the navigation section

GeorgeKerscher:
While I was at it, I tried to address another issue with the language in the navigation. It was aroundthe issue submitted, 469,"Hard to understand navigation.

I reorderd the items toc, index, headings, go to page.

Oh, I  added the term go to page. I know this is a reading system function, but the page list must be there in the content. It just felt like something end users would understand.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -836,7 +836,7 @@
 						adoption.</p>
 
 					<aside class="example" title="Descriptive explanations">
-<p><span data-localization-id="navigation-intro" data-localization-mode="descriptive">Navigation features contains:</span></p>
+<p><span data-localization-id="navigation-intro" data-localization-mode="descriptive">Navigation features contain:</span></p>
 						<ul>
 							<li data-localization-id="navigation-toc" data-localization-mode="descriptive">a table of contents that provides access to all chapters of the text via links</li>
 <li data-localization-id="navigation-index" data-localization-mode="descriptive">An index
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -838,13 +838,13 @@
 					<aside class="example" title="Descriptive explanations">
 <p><span data-localization-id="navigation-intro" data-localization-mode="descriptive">Navigation features contain:</span></p>
 						<ul>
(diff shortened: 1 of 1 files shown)


a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -814,17 +814,15 @@ <h3 data-localization-id="navigation-title">Navigation</h3>
...
+							information about navigation is available</span>.</p>
 				</div>
 
 				<p>Identifies the navigation features included in the publication.</p>
 
-				<p>Offering different ways to navigate digital publications (table of contents, by headings, index,
-					etc.) is important to allow the user to easily access each part of the content. These features are
-					both essential for accessibility and very important for all users reading publications. Therefore,
+				<p>   Navigating a digital publication    through a table of contents, an index, by headings,  or by direct go to page shortcuts is important to allow the user to easily access each part of the digital publication. These features are
+					both essential for accessibility and  important for all users reading publications. Therefore,
```
mattgarrish, line 823: The "proper title selection" phrasing sounds awkward. Could this be written instead as: "Therefore, it is important to clearly communicate what navigation features are available so users can decide if a publication will meet their needs."

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -838,32 +836,33 @@ <h4>Examples</h4>
...
+<li data-localization-id="navigation-index" data-localization-mode="descriptive">An index
+								provides links to referenced entries</li>
 							<li data-localization-id="navigation-structural" data-localization-mode="descriptive"
 								>Contains book elements such as headings, tables, etc for structured navigation</li>
-							<li data-localization-id="navigation-index" data-localization-mode="descriptive">Index
-								provides links to item references</li>
-
 							<li data-localization-id="navigation-page-navigation" data-localization-mode="descriptive">A
-								page list enables users to navigate directly to pages from the identified print source
+								page list enables users to directly go to    pages from the identified print source
```
mattgarrish, line 849: Although split infinitives aren't the end of the world anymore, this one stuck out when I read it. I would drop "directly" from the sentence as it doesn't add anything to the meaning.

gregoriopellegrino:
Hi George,
can you also see issue #475? In the extended version, sentences begin with "Contains..." others with "An index..." or "A page list".

Also repeating "Contains" with every line seems a bit redundant to me.

clapierre (approved):

avneeshsingh:
Yes, we should avoid example that indicate presence of sound hazards.

If we want to give an example of no sound hazard, it is ok because publication without any audio will not have sound hazard.

Issue #471, opened 2024-11-12
Editorial Feedback for the ONIX/EPUB Techniques Document


wareid:
My colleagues at Kobo and I put together this feedback after reviewing the document. Much of this is editorial, with the goal of improving the readability and usability of the documents. There is also some functional feedback too. 

[1. ONIX Metadata](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/techniques/onix-metadata/#onix-metadata) / [1. EPUB Accessibility Metadata](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/techniques/epub-metadata/#epub-accessibility-metadata)
* The entire section is in Notes, which is a bit confusing and makes the content difficult to parse. I'd recommend removing the notes and structuring the content as an introduction more properly. 

[4.3 Conformance](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/techniques/onix-metadata/#conformance-group) / [4.3 Conformance](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/techniques/epub-metadata/#conformance-group)
* This feels incredibly long and convoluted, and I don’t think accepting A, AA, AAA on their own is informative enough to the user, at Kobo we’ve adopted validation rules that throw out levels if there is no attached version, as it’s not informative enough to what they are conformant to.
* For the EPUB metadata, this also requires additional logic to parse and split the value provided in `dcterms:conformsTo` which while not difficult, requires additional processing that seems unnecessary.

[4.5 Navigation](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/techniques/onix-metadata/#navigation) / [4.10 Additional Accessibility Information](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/techniques/onix-metadata/#additional-accessibility-information)
* The structure of these sections feels disconnected with the recommendations made in the UX Guidelines document, I wonder if they need to be reconciled structurally? It would be good to represent the information in a common order. 
* There was similar structural issues in the EPUB Metadata document. 

It was also observed that it would be helpful to review the content of the three documents and ensure they align to one another structurally and content-wise. Mentions of metadata properties in the ONIX and EPUB metadata docs are not made in the main document, which adds to the confusion.

GeorgeKerscher:
Can you expand on the bullet in conformance. You write:
[...]

PR #494, opened 2024-11-25
exact string matching

clapierre:
now making sure the exact strings for conformance are checked

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -346,37 +346,37 @@
 					</dd>
 					<dt><var>epub_accessibility_1_1</var></dt>
 					<dd>
-                        <p>If true it indicates that the <i>conformsTo</i> contains "EPUB Accessibility 1.1"  (EPUB Accessibility Specification 1.1) is present in the package document, otherwise if false it means that the metadata is not present.</p>
+                        <p>If true it indicates that an approved <i>conformsTo</i> claim contains "EPUB Accessibility 1.1"  (EPUB Accessibility Specification 1.1) is present in the package document, otherwise if false it means that the metadata is not present.</p>
 						<p>This means that the publication conforms with the requirements of EPUB Accessibility Spec 1.1.</p>
 					</dd>
 					<dt><var>wcag_20</var></dt>
 					<dd>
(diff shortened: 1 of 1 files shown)


clapierre:
This addresses the following conformance matching string issue brought up by @wareid in issue
#471 Editorial Feedback for the ONIX/EPUB Techniques Document

GeorgeKerscher:
It looks good to me, but I would like to have others review.

GeorgeKerscher (approved):

gregoriopellegrino (approved):

mattgarrish:
Do we really need to set all these variables? From what I can tell, only the conformance string to display and the wcag level matter in terms of setting statements in the following section.

For example, all of the 1.1 matching strings could be reduced to something like:

<span><b>IF</b> the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="dcterms:<i>conformsTo</i>" and matches(normalize-space(.), 'EPUB Accessibility 1\.1 - WCAG 2\.[0-2] Level [A]+')</code> :</span>
                       <span>
                            <b>THEN</b> <b>LET</b> <var>conformance_str</var> = replace(normalize-space(.), ' - ', ''), 
                            and <b>LET</b> <var>wcag_level</var> = replace(normalize-space(.), 'EPUB Accessibility 1\.1 - WCAG 2.[0-2] Level ', '').

(I don't know if we'd really want to use xslt syntax for setting the variables, but it's just to give an idea that you can process the existing string to get the same value you generate below.)

In any case, if conformance_str is not an empty string, then you'd know you can continue to display conformance information. You'd still use the wcag_level to give the generic statement and conformance_str would contain the specific information rather than having to rebuild almost exactly the same string you just decomposed.

The conformance statement would presumably still be translatable; it would just require an extra step to match the English string that results after removing the dash into whatever the equivalent translation would be (i.e., you wouldn't show every possibility in the techniques document, but maybe note that the value could translate to one of 9 possible permutations of 1.1 identifiers).

clapierre:
@gregoriopellegrino what do you think about @mattgarrish suggestion?
I don't know how I would write it up since the only variable set is the WCAG level and how it would work in the Instructions portion without the other variables being set. I am sure there is a more elegant way to do this but I haven't the time to figure that out.

mattgarrish:
I can open a new PR based on this one to give a better idea of how it might work. I assume this processing is based on the onix metadata, but in that case the information comes decomposed and has to be built up. For the conformsTo string, it seems like a lot of unnecessary work to break it all apart just to put it back together.

PR #495, opened 2024-11-26
Reduce variables and processing of epub conformance metadata

### LINKED ISSUE #511: EPUB A11y 1.0 xpaths missing opening quote
In the conformance section, the current xpaths are like this:

> text() = http://www.idpf.org/epub/a11y/accessibility-20170105.html#wcag-aa"

But there needs to be a quote to start the URLs.

### LINKED ISSUE #512: Extra square bracket in certifier report xpath
The bracket at the end is invalid:

> /package/metadata/meta[@property="a11y:certifierReport"]/text()]

mattgarrish:
This pull request implements the processing change I mentioned in https://github.com/w3c/publ-a11y/pull/494#issuecomment-2500931686

In addition to what I mentioned there, I noticed a couple of other problems with the processing that I've tried to fix:

- For one, there's no value in checking for a random conformsTo statement and then say the conformance can't be determined. It would result in an unknown claim for an epub/a publication, for example. I would stay away from making sense of things we don't know and leave it to future versions of the spec to handle any new conformance claims that need noting. I took out the variable and conformance statement related to this.
- Also, the processing of the metadata should end if a known accessibility conformance statement isn't found (i.e., the first "ELSE" condition). If you don't stop there, you could end up showing certifier metadata related to an unknown claim and you'd always have an empty heading for the conformance statement that's not there. I added the condition "AND end processing of conformance statement" to the ELSE, but maybe there's other wording you'd want to use here.

We'd probably still need a note where I have the display statement for conformance_str. I don't know how you guys would want to handle that the variable could hold any of the nine statements. It probably requires a different set of strings than the ONIX processing uses, although the end result will be the same strings.

I'm also open to other ways to process the 1.1 strings into the two variables we need. Maybe we just drop the normalize-space call, as otherwise it'd look like any language's string replacement functionality.

Fixes #511 
Fixes #512 

***

[Preview](https://raw.githack.com/w3c/publ-a11y/simplify-epub-conf/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/simplify-epub-conf/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html)

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -421,7 +421,7 @@
 					
                     <li><b>LET</b> <var>certification_date</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="dcterms:date" and @refines=//meta[@property="a11y:certifiedBy"]/@id]/text()]</code>.</li>
 					
-                    <li><b>LET</b> <var>certifier_report</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="a11y:<i>certifierReport</i>"]/text()]</code>.</li>
+                    <li><b>LET</b> <var>certifier_report</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="a11y:<i>certifierReport</i>"]/text()</code>.</li>
 				</ol>
 				
 				<h4>Instructions</h4>
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -473,18 +473,18 @@
 											<p>The value of <var>conformance_string</var> will be one of the following
 												strings:</p>
 											<ul>
-												<li><code id="conf-epub10-wcag20a">EPUB Accessibility 1.0 WCAG 2.0 Level A</code></li>
-												<li><code id="conf-epub10-wcag20aa">EPUB Accessibility 1.0 WCAG 2.0 Level AA</code></li>
-												<li><code id="conf-epub10-wcag20aaa">EPUB Accessibility 1.0 WCAG 2.0 Level AAA</code></li>
-												<li><code id="conf-epub11-wcag20a">EPUB Accessibility 1.1 WCAG 2.0 Level A</code></li>
-												<li><code id="conf-epub11-wcag20aa">EPUB Accessibility 1.1 WCAG 2.0 Level AA</code></li>
-												<li><code id="conf-epub11-wcag20aaa">EPUB Accessibility 1.1 WCAG 2.0 Level AAA</code></li>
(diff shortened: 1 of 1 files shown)


clapierre:
Ok, Reviewing the original code and the updated code, I se an issue which we didn't really address.  In both cases we never did say what this EPUB has been checked against, only if it met the min, met the expected or exceeded.  

We don't say if it met WCAG 2.0 2.1 or 2.2.  Which is a problem since some folks will require WCAG 2.1 AA, where others might require WCAG 2.2 AA. We also don't point out if this EPUB passed EPUB 1.1 or 1.0.  Not sure if that is as big a deal as the WCAG version or not.

@GeorgeKerscher @gregoriopellegrino what do you all think? Or do we need to discuss this at our next editors call?

GeorgeKerscher:
We have this in:
Permalink for Section 4.3.2
The following information can be placed in a section that shows the details of the conformance information.
Conformance Statement
Identifies the accessibility specification and the conformance level to which the publication assertions are made. When the publication claims to conform to more than one specification, additional conformance statements may be provided.
Certification date
If the date of the certifier's evaluation is provided, then this would be of interest. This is normally associated with the certifier.
Certifier's Report
If a link to a report is provided, this may be of interest.

Perhaps we should use the word must or strongly advise.

mattgarrish:
What we learned with publication manifest is that when making algorithms it's really helpful to write some public code to implement the instructions. That way anyone can run some samples through and see what results. I can maybe try and adapt the generator I made to implement the current logic.

gregoriopellegrino (approved):

clapierre (approved):

mattgarrish:
I made a few changes in the last commit based on implementing this code:

- I renamed `conformance_str` to `conformance_string` to be consistent with naming elsewhere
- I created a variable called `conformance` and use the language for the certifier info to extract the node rather than test that it is present. This makes it simpler to understand the replace functions that extract the statement and wcag level
- I reorganized the instructions so that there isn't the need for that halting statement in the middle. If conformance_string is empty, then you can bail out right away. The processing now happens in the ELSE that follows.

I also added the missing quotes to the 1.0 xpaths and deleted the extra backet in the certifier xpath so this will also close #511 and #512

mattgarrish:
The one thing that make me unsure about merging this yet is the string that results from extracting the 1.1 conformance statement and using a regex to delete the hyphen from it.

The result is that we can't associate a single ID with it. It could be any of the nine strings. Should we list all of these with their IDs or put them in a note? It makes localization a little more complicated as you have to translate the English text to an ID before you can match the localization. It's not like the other instructions that end in an single ID value.

mattgarrish:
I've added a note with the possible values of `conformance_string`. I made up IDs for each as I don't know if these are already handled somewhere or if they need to be added. I won't merge this until I know for sure what these should be.

gautierchomel (approved):

Issue #474, opened 2024-11-13
Make image descriptions more visible in accessibility metadata display


gregoriopellegrino:
While reviewing the second draft of the Accessibility Metadata Display Guide 2.0, we noticed that alternative descriptions for images aren't explicitly visible in the key information sections. Currently, this important accessibility feature is embedded within the broader "Supports nonvisual reading" category.

The presence of alternative descriptions is crucial information both for end users and as recognition of the significant work publishers put into creating them - it's one of the most important and time-consuming tasks when producing accessible publications.

We propose expanding the scope of the current "Charts, diagrams, math, and formulas" section, which is currently quite limited and STEM-focused content. We could rename it to "Non-textual content" (or find a better title) and use it to communicate the presence of:
- Short alternative descriptions for images
- Extended descriptions for complex images 
- Accessible mathematical and chemical formulas
- Charts and diagram descriptions

clapierre:
I love this idea Gregorio!

gautierchomel:
Editor's proposes "visual and complex contents".

GeorgeKerscher:
Closing because image alt text and extended descriptions have been elevated. If something is still missing, please file a new issue.

PR #498, opened 2024-11-27
Added images and photographs  and alt text in nonvisual reading section

GeorgeKerscher:
Simple improvement. alt text is now the first item one encounters.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -480,10 +480,10 @@
 
 				<p>This field answers whether nonvisual reading is possible, not possible, or unknown.</p>
 
-				<p>Digital publications with essential content included in non-textual form (such as graphs, tables or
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I believe this fixes #474

gregoriopellegrino:
The change seems to me to be in the paragraph that explains the key accessibility information, but not of the examples. If we want to show it to end users I think it should be included in the examples.

GeorgeKerscher:
I agree that having this in the examples would emphasise  the importance of alt text and make publishers feel better about all the work they have put into this feature.

We have in the techniques:
from ONIX
• code 14 of codelist 196 (Short alternative textual descriptions);
From EPUB metadata
• accessibilityFeature="alternativeText" (Short alternative textual descriptions);

It may also be possible to derive alt text from accessModeSufficient, but we can discuss this.

I will add an ID into the guidelines when I insert this.

The id will be: has-alt-text

For descriptive: Has alt text on images 
for compact: has alt text

We could go further and do the same for extended descriptions, but we would also have this in complex content, the duplication of information would not be good. I think better to have extended descriptions in complex content.

PR #516, opened 2024-11-28
Changed pre-recorded to prerecorded and removed white space

### LINKED ISSUE #500: Use "prerecorded"
Minor nit, but the word isn't need typically hyphenated, similar to how nonvisual is not.

GeorgeKerscher:
Hi Matt,

I changed pre-recorded to prerecorded in the body, but did not change the IDs, which also had this misspelling.

While I was at it, I removed extra spaces in the content and trimmed trailing spaces.

I am just asking you to give it a quick check. I will need to merge this before I do more work, or I expect that the many changes in white space would trigger conflicts.
I used a regular expression and I think it worked fine.

fixes #500

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -1,1388 +1,2106 @@
 <!DOCTYPE html>
 
 
-<html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
-	<head>
- <meta charset="utf-8" />
- <title>Accessibility Metadata Display Guide for Digital Publications 2.0</title>
- <script src="https://www.w3.org/Tools/respec/respec-w3c" class="remove" defer="defer"></script>
- <script class="remove">
- // <![CDATA[
- var respecConfig = {
- group: "publishingcg",
- specStatus: "CG-DRAFT",
- noRecTrack: true,
- latestVersion: "https://www.w3.org/publishing/a11y/UX-Guide-metadata/principles/",
- edDraftURI: "https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/guidelines/",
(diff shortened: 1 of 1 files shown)


mattgarrish:
> While I was at it, I removed extra spaces in the content and trimmed trailing spaces.

I think it looks good, but it's hard to tell because every line of the file has changed. I only found pre-recorded with a hyphen in some IDs, which seems fine.

Did you really intend to remove all the markup indentation, though? The entire file is left aligned now. If it's easier to work that way, it's fine by me, but most editor apps will start trying to re-indent the markup.

GeorgeKerscher:
Hi,

No, I did not intend to remove the indentation. I guess the search for spaces removed the tabs as well, sigh.

I will try to fix it and commit. If you donot see improvement, I can trash the PR. I just wanted to get rid of the extra spaces in the document, which is really hard to identify with a screen reader.

I will make another commit.

Best
George
 not

GeorgeKerscher:
Hi Matt,

I see all the indention, hopefully it looks OK. I realize that each time I open a file doing something like this would totally mess up the diff. Just trying to get it back to a useable state.

If there is a better way to auto indent to make the  code easier for sighted users, please have at it.

mattgarrish:
Looks better now. My one worry was that by removing all the indentation it might have messed up any examples. Regex replaces won't pay attention to context, although I'm not sure there are any code examples in the guidelines.

> If there is a better way to auto indent to make the code easier for sighted users, please have at it.

One thing I've seen done elsewhere is to run a weekly github action to reformat files. That way it doesn't matter what indentation anyone uses, at it will get standardized after the fact. As you're the primary editor of this document, I don't think we need to worry about it so much, though. Whatever is easiest for you to edit should be fine.

GeorgeKerscher:
I made changes to VS Code and now it will not insert an extra space here and there. I should not need to do the whit space removal, hopefully.

Issue #505, opened 2024-11-28
Charts section results in conflicting statements


mattgarrish:
Because this section uses if statements throughout, it's possible that you can mix statements about accessible content with the final statement saying the accessibility isn't known. All that has to happen is one of the OR conditions in the last statement has to be true. For example, if you have accessible charts but inaccessible math, you'll get both:

> Charts and diagrams have extended descriptions

and

> accessibility of formulas, charts, math, and diagrams not identified as being accessible

mattgarrish:
Closing this as superseded by later changes.

PR #520, opened 2024-11-30
Rewrite of Complex content was Charts diagrams etc

GeorgeKerscher:
I could not find in the issues a new name for this section. From my memory of the discussion, we said "Complex content" so that is what I went with.

I have math chemistry, charts, diagrams, figures, and graphs as the list of items in addition to math and chemistry.

In the examples, I have appended to the IDs we use

For the math:
-as-mathml
-as-latex
-as-descriptive
For chemistry:
-as-mathml
-as-latex

Question:

Should we use the word "mathematics" in the normal text and in the descriptive  text? Some places use maths and others math, but mathematics is the formal usage.
 
-

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -1302,31 +1302,21 @@
 			</section>
 		</section>
 		<section id="charts-diagrams-and-formulas">
-			<h3 data-localization-id="charts-diagrams-formulas-title">
-				Charts, diagrams, math, and formulas</h3>
+			<h3 data-localization-id="charts-diagrams-formulas-title">Complex content</h3>
 
 			<div class="note">
 				<p>This key information can be hidden if metadata is
 					missing. Alternatively it can be stated that
 					<span
 						data-localization-id="charts-diagrams-formulas-unknown"
-						data-localization-mode="descriptive">accessibility
-						of formulas, charts, math, and diagrams not
-						identified as being accessible</span>
+						data-localization-mode="descriptive">accessibility of complex content not identified as being accessible</span>
 				</p>
 			</div>
 
-			<p>Indicates the presence of formulas (including math,
-				chemistry, etc.), graphs, charts, and diagrams
-				within the title and whether these are in an accessible
-				format or available in an alternative form
-				(e.g., whether formulas are navigable with assistive
-				technologies, or whether extended descriptions
-				are available for complex images).</p>
+			<p>Indicates the presence of math, chemistry, charts, diagrams, figures or graphs within the title and whether these are in an accessible format or available in an alternative form, e.g., whether math and chemistry is navigable with assistive technologies, or whether extended descriptions are available for complex images.</p>
 
 			<p>This group should be displayed only if the metadata
-				indicates the presence of formulas, math, charts, or
-				graphs within the title, otherwise it can be hidden.</p>
+				indicates the presence of math, chemistry, charts, diagrams, figures, or graphs within the title, otherwise it can be hidden.</p>
 
 			<section id="cdf-examples">
 				<h4>Examples</h4>
@@ -1337,31 +1327,32 @@
 
 				<aside class="example" title="Descriptive explanations">
 					<ul>
+<li><span data-localization-id="charts-diagrams-formulas-accessible-math-as-mathml"
+								data-localization-mode="descriptive">Math as MathML</span>
+						</li>
+<li><span data-localization-id="charts-diagrams-formulas-accessible-math-as-latex"
+								data-localization-mode="descriptive">Math as Latex</span>
+						</li>
+						<li><span
+								data-localization-id="charts-diagrams-formulas-accessible-math-described"
+								data-localization-mode="descriptive">Math as images with text description</span>
+						</li>
+<li><span data-localization-id="charts-diagrams-formulas-accessible-chemistry-as-mathml"
+								data-localization-mode="descriptive">Chemistry is in MathML</span>
+						</li>
+<li><span data-localization-id="charts-diagrams-formulas-accessible-chemistry-as-latex"
+								data-localization-mode="descriptive">Chemistry is in LaTex</span>
+						</li>
 						<li>
 							<span
 								data-localization-id="charts-diagrams-formulas-extended"
-								data-localization-mode="descriptive">Charts
-								and diagrams are present and described
-								by extended descriptions</span>
-						</li>
-						<li><span
-								data-localization-id="charts-diagrams-formulas-accessible-math"
-								data-localization-mode="descriptive">Contains
-								math formulas in accessible
-								format</span>
-						</li>
-						<li><span
-								data-localization-id="charts-diagrams-formulas-accessible-chemistry"
-								data-localization-mode="descriptive">Chemistry
-								is in an accessible format</span>
+								data-localization-mode="descriptive">Charts and diagrams are present and have extended descriptions</span>
 						</li>
 						<li>
-							<span
-								data-localization-id="charts-diagrams-formulas-unknown"
+							<span data-localization-id="charts-diagrams-formulas-unknown"
 								data-localization-mode="descriptive">
 								accessibility of formulas, charts, math,
-								and diagrams not identified as being
-								accessible</span>
+								and diagrams not identified as being accessible</span>
 						</li>
 					</ul>
 				</aside>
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
As a proposal for statements I would suggest:

```
Contains:
- complex images are described by extended descriptions
- math formulas in accessible format (MathML)
- math formulas in accessible format (LaTex)
- math formulas in accessible format (formulas as images with textual description)
- chemical formulas in accessible format (MathML)
- chemical formulas in accessible format (LaTex)
```

GeorgeKerscher:
I assume the suggestions made by Gregorio is for the descriptive statements. I will make those changes.

I will have the descriptive reference Charts, diagrams,figures, and graphs as complex images in the compact. It will be writen out in descriptive.

I also will order the items as the three math followed by the two chemistry, and then the complex images.

GeorgeKerscher:
in the latest commit, we have math formulas in the examples. In the description we only use the word math. In the compact we have only math. Do we need formulas for the math?

Also I have removed the word chemistry everywhere in this section. It is replaced with chemical formulas. As far as I know, chemical formulas can be represented in MathML and in LaTex. The rest of chemistry is represented in images, so I think this change is technically correct.

mattgarrish:
Is using "complex content" as a title a good idea? It doesn't really convey anything to the average user, and the content itself isn't necessarily complex. Just because there's math doesn't make the equations complex, for example.

The grouping seems like it's for the accessibility of highly visual content. Is there some name more in that vein that would work?

Note that I'm not suggesting we go back to having all the content types in the title again, but try to get closer to what groups these together.

gautierchomel:
[Preview](https://rawcdn.githack.com/w3c/publ-a11y/95733fc769ad6c25c9d3c64e8bd9125d59bf4c33/a11y-meta-display-guide/2.0/draft/guidelines/index.html#charts-diagrams-and-formulas)

gautierchomel:
With apologies, I wanted to suggest changes, but I ended up amending the PR with an extra commit. 

The changes are fixing a repetition: 

- "accessibility of complex content not identified as being accessible" becomes "complex content not identified as being accessible"
- "accessibility of math, chemical formulas, charts, diagrams, figures, and graphs not identified as being accessible" becomes "math, chemical formulas, charts, diagrams, figures, and graphs not identified as being accessible"

gautierchomel (approved):

gautierchomel:
> Is using "complex content" as a title a good idea?

I agree with @mattgarrish that it is not very understandable (isn't a psychcognitive book with text only a complex content in itself?)

Maybe it is only "Contents"

mattgarrish:
Even saying "complex images" is a bit misleading. We don't know if the images are complex or not, as that's not solely what determines whether an image needs a description. It's the amount of information that needs to be conveyed. You could have a chart that isn't terribly complicated to read, for example, but contains a lot of information.

Complex is more of a value judgement.

Looking at this again this morning, couldn't the math and chemistry statements fit under a STEM heading? And is there a way to add image descriptions to the nonvisual reading section? Maybe like how conformance is done with a general statement first and then detailed information like the presence of extended descriptions and transcripts after? As it is, you have to go to two separate sections to find those other pieces of information.

gautierchomel:
The french implementation says "Images described in read aloud and braille" instead of "All content can be read as read aloud speech or dynamic braille"

GeorgeKerscher:
I like those changes.

GeorgeKerscher:
The Complex content section could be renamed to Science, Technology, Engineering, and Math (STEM).

I think we would also want to duplicate the information about extended descriptions under Support For nonvisual reading. In this way, it would only appear twice if there was math or chem. it would follow alt text.

Would this work?

avneeshsingh:
Indeed, we started pondering for changing name because we wanted this heading to go beyond STEM content.

mattgarrish:
> and are very useful for example for users with dyslexia.

How do you know that someone with dyslexia can read an extended description, though? It assumes it's not hidden and linked up through aria-describedby. The only thing we can be reasonably sure of is that it will be available to AT.

You can still separate formulas and equations from image descriptions with literally using stem as the header. I'm still not seeing what connection they have to image descriptions.

mattgarrish:
> 1. To include a area with the heading of , "Mathematical and scientific notation"

I'm good with this.

gregoriopellegrino:
I propose to discuss this in the editors' call, it seems to me that different interpretations and meanings are being mixed up.

GeorgeKerscher:
Gregorio, should I do a commit with that heading to facilitate the discussion tomorrow?

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -1302,21 +1302,21 @@ <h4>Display techniques for navigation support</h4>
...
-			<h3 data-localization-id="charts-diagrams-formulas-title">Complex content</h3>
+			<h3 data-localization-id="charts-diagrams-formulas-title">Rich content</h3>
 
 			<div class="note">
 				<p>This key information can be hidden if metadata is
 					missing. Alternatively it can be stated that
 					<span
 						data-localization-id="charts-diagrams-formulas-unknown"
-						data-localization-mode="descriptive">accessibility of complex content not identified as being accessible</span>
+						data-localization-mode="descriptive">accessibility of rich content not identified as being accessible</span>
```
mattgarrish, line 1312: "accessibility of rich content not identified as being accessible"

Shouldn't this text also be changed to match the change below? It should probably be: "rich content not identified as being accessible"

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -1350,10 +1350,16 @@ <h4>Examples</h4>
...
+Videos included in publications have closed
+							captions </span></li>
+
+<li><span data-localization-id="additional-accessibility-information-adaptation-open-captions" data-localization-mode="descriptive">
+Videos included in publications have open captions </span></li>
+<li>
 							<span data-localization-id="charts-diagrams-formulas-unknown"
 								data-localization-mode="descriptive">
-								math, chemical formulas, charts, diagrams, figures, and graphs not identified as being accessible</span>
+								math, chemical formulas, charts, diagrams, figures, graphs, and videos not identified as being accessible</span>
```
mattgarrish, line 1362: What happens if only some of these things are not identified as being accessible? I wouldn't think it'd happen too often that they're all found in a book and they're all not identified as accessible. This relates to the problem with the techniques that I noted in #505 that the statement about all rich content types not being identified could be output at the same time as statements about individual rich content being accessible.

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -1302,21 +1302,21 @@ <h4>Display techniques for navigation support</h4>
...
 				</p>
 			</div>
 
-			<p>Indicates the presence of math, chemical formulas, charts, diagrams, figures or graphs within the title and whether these are in an accessible format or available in an alternative form, e.g., whether math and chemical formulas are navigable with assistive technologies, or whether extended descriptions are available for complex images.</p>
+			<p>Indicates the presence of math, chemical formulas, charts, diagrams, figures or graphs within the title and whether these are in an accessible format or available in an alternative form, e.g., whether math and chemical formulas are navigable with assistive technologies, or whether extended descriptions are available for complex images. In addition, it indicates the presence of videos and if closed captions or open captions are available.</p>
 
 			<p>This group should be displayed only if the metadata
-				indicates the presence of math, chemical formulas, charts, diagrams, figures, or graphs within the title, otherwise it can be hidden.</p>
+				indicates the presence of math, chemical formulas, charts, diagrams, figures, graphs, or videos within the title, otherwise it can be hidden.</p>
```
mattgarrish, line 1319: Weren't we going to add audio and video? I'm only finding video added.

GeorgeKerscher:
We have the section on audio books. That is more or less the whole book as audio and with synced media.

In the Rich content section, I am adding portions of audio content  and transcripts. In my mid this would cover things like a reading of a poem and other short clips.

There may be something to change in the techniques; there may be two variable conditions associated with the ID I am using.

We do not have descriptive audio as a metadata value anywhere? I have seen it in videos, but it is not common outside of movies.

 joined

mattgarrish:
> In the Rich content section, I am adding portions of audio content and transcripts. In my mid this would cover things like a reading of a poem and other short clips.

Right, in the text I clicked to comment on it has this:

> indicates the presence of math, chemical formulas, charts, diagrams, figures, graphs, or videos within the title

With videos being the new addition. I was expecting it would have audio and video listed in the string if we're capturing audio clips here. I think the same applies to a couple of other spots where only video is mentioned.

GeorgeKerscher:
I have the commit in process. Should have it done later.

PR #536, opened 2024-12-13
From dec 12 discussion

GeorgeKerscher:
Oops, forgot pull before I made changes, sigh. May need to throw this away. I will see.

Made the capitilization changes Matt suggested.
Added No information is available to the end of the examples. Where the logic was there, just inserted. It was not there in several places. These do not have an ID in the paragraph or list item.

We need to take a close look at the logic in te techniques for the Rich content section. I see several problems.


clapierre (approved):

GeorgeKerscher:
I think we only have a problem where there are features identified being present. So the diagram on visual chemistry on visual,  essentially all the visual presence. The images being present is not an issue, because that is addressed separately in the alt text section.

I don't know if the presence of videos is identified separately from having closed and open captions.

PR #543, opened 2024-12-17
Changes made that Gregorio, Charles, and George agreed on

GeorgeKerscher:
Gregorio, Charles, and George met today. We discussed several items covered in this PR
1. In the navigation compact examples, the first letter was made uper case.
2.  In the Rich content section, the one statement was removed that said that charts, diagrams, figures, and graphs werenot identified as being accessible.
3. Added to the introduction of key information that the order and the heading used to show is up to the implementation. The headings is the key item here, because we expect implementers to have several groups under a single heading, such as "Reading Modes."

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -551,8 +551,7 @@
 			<div class="note">
 				<p>This document does not define the order in which to
 					show the key accessibility information; each
-					implementer can decide the preferred order for
-					showing the accessibility information that
+					implementer can decide the preferred order and appropriate headings to use to show the accessibility information that
 					follows.</p>
 				<p>In this document are examples of important metadata
 					that is expected to be in a wide range of
@@ -1233,21 +1232,19 @@
 
 					<ul>
 						<li><span data-localization-id="navigation-toc"
-								data-localization-mode="compact">table
-								of
-								contents</span></li>
+								data-localization-mode="compact">Table
+								of contents</span></li>
 						<li><span
 								data-localization-id="navigation-index"
-								data-localization-mode="compact">index</span>
+								data-localization-mode="compact">Index</span>
 						</li>
 						<li><span
 								data-localization-id="navigation-structural"
-								data-localization-mode="compact">headings</span>
+								data-localization-mode="compact">Headings</span>
 						</li>
 						<li><span
 								data-localization-id="navigation-page-navigation"
-								data-localization-mode="compact">go to
-								page </span></li>
+								data-localization-mode="compact">Go to page </span></li>
 <li>No information is available</li>
 					</ul>
 				</aside>
@@ -1316,7 +1313,7 @@
 						<li>
 							<span
 								data-localization-id="charts-diagrams-formulas-extended"
-								data-localization-mode="descriptive">Charts, diagrams, figures, and graphs have extended descriptions</span>
+								data-localization-mode="descriptive">Complex images are described by extended descriptions</span>
 						</li>
 <li><span data-localization-id="additional-accessibility-information-adaptation-closed-captions" data-localization-mode="descriptive">
 Videos included in publications have closed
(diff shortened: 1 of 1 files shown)


clapierre (approved):

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -551,8 +551,7 @@ <h3>Introduction to key accessibility information</h3>
 			<div class="note">
 				<p>This document does not define the order in which to
 					show the key accessibility information; each
-					implementer can decide the preferred order for
-					showing the accessibility information that
+					implementer can decide the preferred order and appropriate headings to use to show the accessibility information that
```
gautierchomel, line 554: Not sure then if it makes sense to include headings in the localisation file as it would introduce confusion (i.e. use only part of the localisation and provide a second localisation to refine implementation with our own heading structure).

a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -1316,7 +1313,7 @@ <h4>Examples</h4>
 						<li>
 							<span
 								data-localization-id="charts-diagrams-formulas-extended"
-								data-localization-mode="descriptive">Charts, diagrams, figures, and graphs have extended descriptions</span>
+								data-localization-mode="descriptive">Complex images are described by extended descriptions</span>
```
gautierchomel, line 1316: A simple photography can be extendly described because of the book context or publisher will. 
I suggest we stick here with the minimum common denominator: images are described by extended descriptions
(I remember we had  a previous discussion about qualifying things as complex but can't find the reference)

gautierchomel (changes requested):
I suggest to keep _Images_ only instead of _Complex Images_ as we have no mean to qualify the complexity of an image. 

Also, the Heading structure decision to implementers (due to our heading structure being problematic) questions the presence of those headings in the localisation file. I will open a separate issue to discuss that.

GeorgeKerscher:
We have elsewhere that images have alt text. Here we are talking about rich content. Having complex images with extended descriptions differenciates images that are sufficiently addressed with simple alt text.

a11y-meta-display-guide/2.0/draft/guidelines/index.html
clapierre: @gautierchomel I brought up this same question if it should just be "images" instead of "complex images" however regular simple images do not need anything other than an alt text description. This is why we chose "complex images" is there a better word to describe images needing more than just a simple alt text description?

gautierchomel: I disagree, a content creator can choose to add an extended description on a "simple" image, because of the context or will to see it displayed. Nothing prevents him to do so.

a11y-meta-display-guide/2.0/draft/guidelines/index.html
clapierre: Something to consider. The same image say of the "Eiffel Tower" can just be a simple image with alt text "Eiffel Tower" but in an Structural Engineering textbook could have that same image but need a detailed description describing how the support beams are connecting, in this case the "simple image" becomes a "complex image" since it requires the extended description.  This is how I look at this anyway.  So it doesn't really matter how "simple" or how "complex" the image is just in our thinking that simple images that don't require additional information are simple and those that require or that the author feels they need to further describe with a longer description we classify as "complex" but this word can be whatever we wish to call any image having an additional description.

mattgarrish:
I'm sort of ambivalent about "complex images" only because you can find a definition of the term from WAI that largely matches the use here: https://www.w3.org/WAI/tutorials/images/complex/

We had this discussion with using the complex in the old "complex content" title about what we're saying about the nature of a book. Complexity of the images itself really isn't the issue here, it's the amount of information that has to be conveyed. Will end users understand that when we say complex images are described it doesn't mean that less complex images may be inaccessible?

gautierchomel (approved):
Nice wording

PR #544, opened 2024-12-17
Added a note about the page break source to navigation before examples

### LINKED ISSUE #484: Pagination source not displayed
EPUB Accessibility 1.1 states:

> Users need to know the source of the pagination in an EPUB publication to determine whether it will be useful for their needs.
> ...
> When an EPUB publication includes page break markers and/or a page list that correspond to a statically-paginated version of the publication, EPUB creators MUST identify that source in the package document metadata.

For this, the `a11y:pageBreakSource` property exists.

Recently, support for identifying the pagination source for reflowable EPUBs using this property was added to InDesign.

ONIX provides a related code (13 in Codelist 51) to identify that a digital product corresponds to a specific print edition. 

However, neither the Display Guide nor the Techniques mention displaying the pagination source to users.

This raises a question: Why are content creators required to include this metadata, and why do we promote adding support for it in production software, if it is not displayed to users?

GeorgeKerscher:
Added this note about the page break source. Several problems here. First, the pageBreakSource is only an identifier. It would require advanced processing to retreive and present the bibliographic information about the print source.

Second, the system presenting the accessibility metadata may not be able to resolve and retrieve information about the other title referenced by the identifier .

So, it says while not required and if possible...

Iin the example, it does not show the bibliographic information.

This fixes #484

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -1198,6 +1198,8 @@
 						href="https://www.w3.org/TR/WCAG21/#multiple-ways">Success
 						Criterion 2.4.5 Multiple
 						Ways</a></cite>).</p>
+<div class="note"><p>While not required if available, it may be helpful to identify the bibliographic information about the page break source identified in metadata of the digital publication. Having sufficient information to identify the print source would assure students and educators that the digital version they are using is correctly associated with the print version used in the classroom. The goal is to have the go to page function take the digital user to the same page as the print version user.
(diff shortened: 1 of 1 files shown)


gautierchomel (approved):

clapierre (approved):

gautierchomel:
FYI, we opened a discussion to provide a means to identify print page sources in Thorium Reader: [Thorium provides no mean to identify print page source #2726](https://github.com/edrlab/thorium-reader/issues/2726)

gregoriopellegrino (approved):

PR #547, opened 2024-12-19
ONIX conformance: LIA compliant

gregoriopellegrino:
While working on the implementation we realized that in the Conformance part of ONIX we did not provide for files declared "LIA compliant," these are legacy content that LIA has certified (the program started over 10 years ago, before EPUB Accessibility 1.0).

These files appear to be certified by LIA and are primarily Italian.

An early proposal of mine can be found here. I would like to discuss this in the editors' call.


clapierre (approved):
There is no equivalent I believe in EPUB Schema.org, this is fine here for ONIX techniques.

GeorgeKerscher:
These files conform to WCAG 2.0 at the AA level, right? The point here is that the files were certified prior toEPUB Accessibility Conformance and Discovery specification 1.0 and therefore have no strings to compare to, is that correct? Good to talk through on the editors call.

clapierre:
Ok to merge per Gregorio's email and our call this morning.

PR #553, opened 2024-12-23
Small fixes for localization

gregoriopellegrino:
I am working on the three documents together: guidelines, ONIX techniques, EPUB metadata techniques.

I started the synchronization work and realized several small problems, I solved those that were no-brainers, but there are some tasks that requires more work or a discussion:
- string alignment "No information is available" in EPUB metadata techniques
- fixing IDs in the Conformance section in the guidelines
- fixing IDs in the Rich contents section in the guidelines
- lack of "prerecorded-audio-complementary" in the guidelines
- not sure about the string "Audiobook" for "prerecorded-audio-only", at least not from ONIX's point of view

Finally in the EPUB Accessibility Metadata techniques are to be fixed, to align with the guidelines: removal of headings and conformance section.

Upon completion of this work I would merge and then proceed to create the draft canonical JSON with the script.


GeorgeKerscher:
Hi,

I was thinking that  we would not need to show the following or I could put it into additional information.
• lack of "prerecorded-audio-complementary" in the guidelines

In my mind this would be a small snip of audio, like the reading of a poem. It is hard to explain and hard to know what is there.

gautierchomel (approved):

PR #556, opened 2024-12-26
Made two changes in guidelines to align with ONIX techniques

GeorgeKerscher:
Very simple changes to align with the Excel file in the ONIX column.

There is still an item in conformance where I think the guidelines are correct. Is when there is a conformance statement, but not to an accepted standard.


avneeshsingh (approved):

gregoriopellegrino:
> There is still an item in conformance where I think the guidelines are correct. Is when there is a conformance statement, but not to an accepted standard.

This is correct, but in ONIX you cannot express it, while you have to keep the string for EPUB techniques.

gregoriopellegrino (approved):

PR #582, opened 2025-01-07
Remove invalid use of normalize-space

mattgarrish:
Sorry, gave some bad advice earlier. normalize-space(.) won't work to select nodes. We can just strip these since epub meta elements will only ever have text.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -411,7 +411,7 @@ dcterms:conformsTo" and normalize-space() = "http://www.idpf.org/epub/a11y/acces
                     </li>
 
                    <!-- EPUB Accessibility 1.1 WCAG 2.X (A/AA/AAA) -->
-					<li><b>LET</b> <var>conformance</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="dcterms:<i>conformsTo</i>" and matches(normalize-space(.), 'EPUB Accessibility 1\.1 - WCAG 2\.[0-2] Level [A]+')]/normalize-space(.)</code>.</li>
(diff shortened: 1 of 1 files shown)


a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -411,7 +411,7 @@ <h4>Variables setup</h4>
                     </li>
 
                    <!-- EPUB Accessibility 1.1 WCAG 2.X (A/AA/AAA) -->
-					<li><b>LET</b> <var>conformance</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="dcterms:<i>conformsTo</i>" and matches(normalize-space(.), 'EPUB Accessibility 1\.1 - WCAG 2\.[0-2] Level [A]+')]/normalize-space(.)</code>.</li>
+					<li><b>LET</b> <var>conformance</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="dcterms:<i>conformsTo</i>" and matches(normalize-space(.), 'EPUB Accessibility 1\.1 - WCAG 2\.[0-2] Level [A]+')]</code>.</li>
```
clapierre, line 414: and matches(normalize-space(.) is still present.

mattgarrish: Ah, that was from the original change. It works in there, but we should definitely remove it for consistency.

mattgarrish:
```suggestion
					<li><b>LET</b> <var>conformance</var> be the value of the node extracted from <var>package_document</var>, using the xpath <code class="xpath">/package/metadata/meta[@property="dcterms:<i>conformsTo</i>" and matches(normalize-space(), 'EPUB Accessibility 1\.1 - WCAG 2\.[0-2] Level [A]+')]</code>.</li>
```

PR #593, opened 2025-01-10
New heading structure added to 2.1

clapierre:


gregoriopellegrino:
My only doubt is referencing H2, because it is very HTML, whereas in my opinion techniques should be agnostic to the technology used for implementation.

Can we include links to the various sections where the code to be inserted under each heading is defined?

clapierre:
Sure, I can just refer to it as a heading,
I will also provide links to the sections for each.

mattgarrish (approved):

gregoriopellegrino (approved):

PR #600, opened 2025-01-17
fixes #599

### LINKED ISSUE #599: EPUB Techniques for pre-recorded audio
Looking over this section I feel there is a problem with the logic.

Couldn't we have a situation where the accessModeSufficient contains "audio" (it may also have another "textual")
and the accessibilityFeature of "sychronizedAudioText"

So in this case if  sychronizedAudioText exists we should always display "Prerecorded audio synchronized with text"

This is currently not what the logic dictates and we would get "Prerecorded audio only" instead.

Do we also need to consider this scenario:

sychronizedAudioText without "audio" as an accessModeSufficient (implies portions of the EPUB will have synchronized audio with text)?

clapierre:


GeorgeKerscher:
Looks nice and clean.

GeorgeKerscher (approved):

gregoriopellegrino (approved):
There doesn't seem to be the same problem on ONIX, for me ok

PR #602, opened 2025-01-20
Create new statements sections for metadata strings

mattgarrish:
Opening this as a draft pull request to get feedback on whether this is a direction we want to go in.

In a nutshell, I found it problematic that we introduce all the strings as examples, and the examples show every string instead of how the metadata is expected to be displayed. I've also always found it hard to have to go to different examples and read the strings to match up the compact and descriptive text.

To address this, I've been discussing with George how we can add new "Statements" section for each technique. The compact and descriptive statements will be grouped and introduced similar to how the techniques work with "if" statements. We can then add real examples of the output for each technique.

So far, I've added all the statement sections, but I still need to add more examples. There is one [example for ways of reading](https://raw.githack.com/w3c/publ-a11y/fix/normative-strings/a11y-meta-display-guide/2.0/draft/guidelines/index.html#va-examples) to show how I'd like to do them.

At any rate, it would help to hear what people think of this change before doing anything more.

***

[Preview](https://raw.githack.com/w3c/publ-a11y/fix/normative-strings/a11y-meta-display-guide/2.0/draft/guidelines/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/fix/normative-strings/a11y-meta-display-guide/2.0/draft/guidelines/index.html)

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -620,7 +620,7 @@
 					clipped by its container).</p>
 	
 				<section id="va-statements">
-					<h5>Statements</h5>
+					<h5>Display statements</h5>
 	
 					<p>The descriptive and compact statements for visual adjustment are as follows.</p>
 					
@@ -717,7 +717,7 @@
 					the nonvisual content.</p>
 	
 				<section id="nv-statements">
-					<h5>Statements</h5>
+					<h5>Display statements</h5>
 	
 					<p>The descriptive and compact statements for nonvisual reading are as follows.</p>
 					
@@ -812,7 +812,7 @@
 					books.</p>
 	
 				<section id="prerecorded-audio-statements">
-					<h5>Statements</h5>
+					<h5>Display statements</h5>
 					
 					<p>The descriptive and compact statements for prerecorded audio are as follows.</p>
 					
@@ -963,7 +963,7 @@
 			</section>
 
 			<section id="conf-statements">
-				<h4>Statements</h4>
+				<h4>Display statements</h4>
 				
 				<div class="note">
 					<p>Some conformance statements incorporate publication metadata. To indicate where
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
I like this approach, it's more simple to understand.

gregoriopellegrino (approved):

GeorgeKerscher:
I have been reviewing this over the weekend. I really like the direction it is taking. I think it will be much easier for people to understand.

Matt askes about getting feedback from people. I think we mean beyond the editors. I also like the idea of examples from the typical of content we expect to see in the real world.

GeorgeKerscher:
I think it is straightforward.  Also, I do my recording for the webinar this afternoon. We could get this version before a larger community on the webinar on Wednesday. A great way to get broad input.

I propose we merge this and we use it asthe basis for my part of the webinar, which is the overview.

If there is agreement, please merge it.

mattgarrish:
I can't approve my own pull request, but for what it's worth I think it's fine to move ahead with merging this and continuing to add more examples after. We're not adding or deleting any information, only rearranging what we had.

@clapierre @avneeshsingh @gautierchomel do you have any concerns?

GeorgeKerscher (approved):

mattgarrish:
> One of the reason for using the term "examples" was to indicate that implementers are free to use another display statement as per their audience.
> Also, we are not providing comprehensive set of statements for covering all the cases.

I think we can do this without defining the strings in example markup. It reads like these strings don't carry any weight at all. Perhaps this belongs more prominently in the introduction.

I can take a stab at consolidating some of the preliminary information and adding a section for this as I think the first three sections are all just introductory material (i.e., they belong together).

GeorgeKerscher:
I think you are correct. It would be good if you take a stab at it. This would make the key information move to 3 instead of 4, or even 2, I think.


George

PR #603, opened 2025-01-20
Update crosscheck strings epub-onix-canonical_json.xlsx

gregoriopellegrino:
I see two major issues:
- I have to update ids for ONIX in ways of reading section
- the new way EPUB techniques manage compliance strings makes them incompatible with the guidelines and ONIX ones


GeorgeKerscher:
As far as I can tell, this looks good to merge.

Charles, will you do this?

PR #610, opened 2025-01-22
Match onix and epub conformance string builds

mattgarrish:
I was worried we were going to have this problem with the conformance string being built from very different metadata.

The pull request aligns the epub conformance string building with the onix so that the strings and IDs will match more readily. It's more processing than the epub strings need for English display, but not knowing if the epub or wcag names need localization it's safer to break them apart and let the strings be built this way.

And in the end, having to have the note with all the possible strings and explaining how to match IDs probably didn't reduce the complexity of the section any.

***

[Preview](https://raw.githack.com/w3c/publ-a11y/fix/conf-strings/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/fix/conf-strings/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html)

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -496,7 +496,7 @@
 
 				<h4 id="conformance-instructions">Instructions</h4>
 				<ol class="condition">
-					<li><span><b>IF</b> NOT <var>epub_version</var> AND NOT <var>wcag_version</var>:</span>
+					<li><span><b>IF</b> <b>NOT</b> <var>epub_version</var> <b>AND NOT</b> <var>wcag_version</var>:</span>
 						<span><b>THEN</b> display <code id="conformance-no">"No information is available"</code>.</span>
 					</li>
 					<li>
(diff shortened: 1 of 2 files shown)


mattgarrish:
Note, I also fixed the string having whitespace at the beginning and end, which when combined would double-pad the spacing.

gregoriopellegrino (approved):
Fine for me, thanks!

mattgarrish:
The last commit just made a couple of changes to the pseudo-code to better follow programming conventions. I changed the '=' string comparison tests to '==' and used $1 for the capture group instead of \1.

mattgarrish:
I've integrated these changes into the viewer and everything is working fine, so it should be okay to merge if everyone is happier with the strings being done this way.

clapierre (approved):

PR #614, opened 2025-01-23
Canonical JSON

gregoriopellegrino:
Hi,
in this PR you will find the first public version of the script for generating canonical JSON from the guidelines and techniques documents.

Also find the updated canonical JSON.

See you later!


clapierre (approved):

gautierchomel (approved):

a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
```diff
@@ -6,269 +6,316 @@
...
+            "compact": "The publication was certified by",
+            "descriptive": "The publication was certified by"
+        },
+        "conformance-certifier-credentials": {
+            "compact": "The certifier's credential is",
+            "descriptive": "The certifier's credential is"
+        },
+        "conformance-certifier-report": {
+            "compact": "For more information refer to the certifier's report",
+            "descriptive": null
```
gautierchomel, line 84: Probably, instead of Null, we should reproduce the Compact.

gautierchomel (approved):

a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
mattgarrish: You can iterate over the json and automatically populate the null values with the compact ones at run time.

Or code the output so that you never attempt to use the descriptive statement in those cases that are null.

Or write some bloated checking to test if you need the descriptive statement and take the compact when it's null.

I can live with null values, in other words, although my preference would be to not have them. If it means we have to duplicate every string in the guidelines so they can be harvested, that might sway me the other way. But could the script that builds the json automatically insert the compact statement instead of null?

gregoriopellegrino: @mattgarrish there is no need to have all the strings in the guidelines, I can fix the script that generates the JSON to use the same value for compact and descriptive, in case there is no descriptive

mattgarrish: I would double up the strings, then, rather than use null.

We're already repeating the strings in the json for the no information available outputs.

a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
```diff
@@ -6,269 +6,316 @@
...
+        "conformance-claim": {
+            "compact": "This publication claims to meet",
+            "descriptive": null
+        },
+        "conformance-details": {
+            "compact": "Detailed conformance information",
+            "descriptive": null
+        },
+        "conformance-epub-accessibility-1-0": {
+            "compact": "EPUB Accessibility 1.0",
```
mattgarrish, line 95: The whitespace needed to join these strings together is missing in the new output.

gregoriopellegrino: Right, I currently use the `normalize-space` function because there are multiple spaces inside the extracted strings. I need to reason how to remove the multiple spaces, keeping the leading and trailing spaces.

a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
```diff
@@ -6,269 +6,316 @@
...
+        },
+        "conformance-certifier-report": {
+            "compact": "For more information refer to the certifier's report",
+            "descriptive": null
+        },
+        "conformance-claim": {
+            "compact": "This publication claims to meet",
+            "descriptive": null
+        },
+        "conformance-details": {
```
mattgarrish, line 90: This is the section heading so I'd expect it to be like the -title headings and not have compact and descriptive versions.

a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
gregoriopellegrino: I think I can find a solution, even when the software reformats the file.

a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
gregoriopellegrino: To handle this we should change the ID to `conformance-details-title`

PR #616, opened 2025-01-23
Restructuring intro and additional accessibility info

clapierre:


clapierre:
This fails color contrast min requirements with only a 2.1:1 instead of the 3.1 for large fonts.
Now that being said this is a Logo which doesn't need to meet color contrast ratios so I am not overly concerned just thought I would let you know that some low vision users may not be able to read the text "Benetech".

GeorgeKerscher (approved):

gregoriopellegrino (approved):

PR #627, opened 2025-01-27
Update conformance examples

mattgarrish:
I was looking to standardize the conformance examples with the other ones I added, but I noticed that they're basically all the same. The only statement that changes is the initial generic conformance one. Otherwise, all the output is identical. Making three examples with compact and descriptive displays would mean we'd have essentially the same info six times.

If we're going to have multiple examples, we should try and make them substantively different.

For now, what I've done is remove the second and third examples that only differed by the "meets minimum" and "exceeds accepted" lines. That leaves one example of meeting accessibility standards in compact and descriptive form and the other about no information being available.

If we want to show the other two strings as examples, maybe we could reduce the amount of information to make them shorter?

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -1144,19 +1144,6 @@
 			<section id="conf-examples">
 				<h4>Examples</h4>
 
-				<p>Four examples are provided for the conformance
-					statement, one shows a statement that claims to
-					meet recommended accessibility standards and a
-					second that claims to meet the minimum level. The
-					third example shows a publication with unknown
-					accessibility, and the final one shows that the
-					conformance information is missing.</p>
-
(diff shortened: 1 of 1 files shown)


mattgarrish:
Links:

***

[Preview](https://raw.githack.com/w3c/publ-a11y/editorial/ex-cleanup/a11y-meta-display-guide/2.0/draft/guidelines/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/guidelines/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/editorial/ex-cleanup/a11y-meta-display-guide/2.0/draft/guidelines/index.html)

PR #629, opened 2025-01-27
Add compact and descriptive conformance statements

mattgarrish:
I just noticed that the compact and descriptive strings for the detailed conformance statement vary in whether they write out WCAG or not. I didn't spot this when I combined the strings in the guidelines, so right now it shows the compact string with some placeholder values.

For now, I opted to write out every string so that they can all be ID'd. (I also fixed up a few bugs on some of the other strings.)

If we don't want to list 12 sets of compact and descriptive conformance strings, the other option would be to use the old approach and use placeholder values. In that case, we'd have the following pair of strings:

 > Compact: This publication claims to meet EPUB Accessibility `<1.X>` WCAG `<2.X>` Level `<A or AA or AAA>`
 > Descriptive: This publication claims to meet EPUB Accessibility `<1.X>` Web Accessibility Content Guidelines (WCAG) `<2.X>` Level `<A or AA or AAA>`

But in this scenario most of the strings wouldn't be taggable with the data-* attributes.

PR #630, opened 2025-01-28
Restructure the conformance display statements

mattgarrish:
I think it would be helpful to separate the general conformance statements that are recommended from the detailed conformance statements, especially since the latter go under a separate heading.

This PR builds on #629 by moving the statements under either a general or detailed conformance subsection of the conformance statements section.

It also gets rid of the explainer subsection as I noticed it was just repeating the compact and descriptive statements and not adding anything new to them. Only the certifier name and credential had any kind of explanation, but I don't think people will get confused by either of those fields.

It also removes the detailed conformance section that was listing the three properties that would go under that heading, since that becomes redundant info when the statements are separated into their own section.

If this is okay, it should not be merged until after #629.

***

[Preview](https://raw.githack.com/w3c/publ-a11y/fix/conf-structure/a11y-meta-display-guide/2.0/draft/guidelines/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/guidelines/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/fix/conf-structure/a11y-meta-display-guide/2.0/draft/guidelines/index.html)

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -966,567 +966,486 @@
 				publication. The certifying organization should be
 				identified along with their credentials and
 				placed immediately after the conformance statement.</p>
-
-
-			<section id="detailled-conformance-information">
-				<h4>Detailed conformance information</h4>
-
-				<p>The following information can be placed in a section
-					that shows the details of the conformance
-					information.</p>
-
-				<dl>
-					<dt>Conformance Statement</dt>
-					<dd>Identifies the accessibility specification and
-						the conformance level to which the
-						publication assertions are made. When the
-						publication claims to conform to more than one
-						specification, additional conformance statements
-						may be provided.</dd>
-
-					<dt id="certificate-date">Certification date</dt>
-					<dd>If the date of the certifier's evaluation is
-						provided, then this would be of interest. This
-						is normally associated with the certifier.</dd>
-
-					<dt id="certifier-report">Certifier's Report</dt>
-					<dd>If a link to a report is provided, this may be
-						of interest.</dd>
-				</dl>
-			</section>
-
+			
 			<section id="conf-statements">
 				<h4>Display statements</h4>
-
+				
 				<div class="note">
 					<p>Some conformance statements incorporate publication metadata. To indicate where
 						this text will be inserted, the statements include placeholder variables. These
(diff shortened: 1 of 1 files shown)


gregoriopellegrino:
Do we need all the examples in the guidelines for all the permutations of EPUB Accessibility version, WCAG version, and WCAG level?
In my opinion in the guidelines we can put an example for each level of WCAG, I would leave the rest to the techniques.

clapierre:
> Do we need all the examples in the guidelines for all the permutations of EPUB Accessibility version, WCAG version, and WCAG level?

I was wondering the same thing.
Couldn't we do the latest Accessibility 1.1 & WCAG 2.2 A/AA/AAA and say in a note that this would be similar for the other verions of Accessibility 1.0, and WCAG 2.0 & 2.1?

mattgarrish:
This is the question I asked in https://github.com/w3c/publ-a11y/pull/629#issue-2813892162

Do we need all the strings for harvesting, since the descriptive versions aren't used in the techniques? If we do, then we need to tag at least one permutation of "Web Content Accessibility Guidelines 2.X" otherwise those strings will never occur in the documents. 

I'd rather use the two placeholders I mentioned in that comment if there's some other way to harvest those strings:

 > Compact: This publication claims to meet EPUB Accessibility `<1.X>` WCAG `<2.X>` Level `<A or AA or AAA>`
 > Descriptive: This publication claims to meet EPUB Accessibility `<1.X>` Web Content Accessibility Guidelines (WCAG) `<2.X>` Level `<A or AA or AAA>`

Worst case, we could sneak only the needed descriptive versions we need into a note after explaining how the placeholders work. For example:

> NOTE:
> The placeholders in these strings allow each version of the respective standard. For example, for WCAG those would be:
> - Web Content Accessibility Guidelines (WCAG ) 2.0
> - Web Content Accessibility Guidelines (WCAG ) 2.1
> - Web Content Accessibility Guidelines (WCAG ) 2.2

clapierre:
I think I like your idea Matt where in a note we can have those other specifications for the string translations.

gregoriopellegrino:
@mattgarrish it's ok for me your last proposal

Issue #631, opened 2025-01-28
Complementary audio and text


mattgarrish:
I thought we were going to drop the statement about embedded audio clips from the ways of reading section, but at the least I think the compact description needs rewriting. The descriptive version says:

> Portions of prerecorded audio are available

That's significantly different from saying there is complementary text for the audio. Looking at the epub techniques, the test only checks if there are audio sources based on there being a primary audio access mode, not whether there is any text that complements it.

clapierre:
Right in EPUB Techniques in the ways to read we have

ELSE IF audio_content:
THEN display "Complementary audio and text"[ID: ways-of-reading-prerecorded-audio-complementary].

Which I think is fine to be in Ways of Reading as this is correct, just means you will have some auditory in the book and there is text which is not synchronized the audio could be anything not really associated with the text. 
However, what should the string be?  "Complementary audio and text" vs. "Portions of prerecorded audio are available".

mattgarrish:
Complementary audio and text sounds like you're saying there are text equivalents. One complements the other. That's the disconnect I have between it and the descriptive which isn't saying anything about whether there is text or not.

What if the strings were something like:

> Prerecorded audio clips
> Prerecorded audio clips are embedded in the content

Is that what we want to say here?

clapierre:
I can see that confusion. I like those two options would one be the short and the other the descriptive version or this was our two options for the simple short description?

mattgarrish:
I was thinking one compact and the other descriptive. I can't think of really different ways of phrasing this.

clapierre:
Work for me, once the Alphabetical PR #624 has been merged I can make the EPUB Technique change to change the string.

GeorgeKerscher:
I agree.  Why don't we make the descriptive and compact the same?

clapierre:
Fixed in EPUB Techniques.

GeorgeKerscher:
Fixed in guidelines-jan-30

HadrienGardeur:
> we're not currently testing whether synchronized text and audio is available for part of the text (`accessModeSufficient` doesn't have `auditory` on its own) or the entire text (`accessModeSufficient` has `auditory` on its own)

Should I open a separate issue about this?

Issue #587, opened 2025-01-08
Different ordering of output in supports nonvisual reading


mattgarrish:
In the onix techniques, the alternative text output is a subtest:

1. IF all_necessary_content_textual:
  THEN display "Readable in read aloud or dynamic braille".
    1. IF textual_alternative_images:
       THEN display "Has alt text". 

In the epub techniques, it occurs before any of the other output

1. IF textual_alternative_images:
  THEN display "Has alt text".
2. IF all_necessary_content_textual:
   THEN display "Readable in read aloud or

I'm not sure which of these is correct. It seems odd to output that there is alternative text before answering if the text is readable, but it doesn't seem right not to say there is alternative text just because the entire content can't be read aloud or read in braille.

mattgarrish:
Does someone want to pick a winner here?

I'd vote for the onix approach of outputting the alt text statement after.

clapierre:
I am fine with that and will submit a PR for EPUB to change its order once we have merged the alphabetical updates.

GeorgeKerscher:
As I recall, because the publishers have put so much into creating alt text, we want to put it front and center.

Also, in my mind I would want to know that images have alt text and then that it is all readable.

So, IMO the EPUB has the correct order.

HadrienGardeur:
The ONIX approach is better IMO.

avneeshsingh:
I think that this is not a high stake decision, I do not have a strong opinion here.
In case it helps in making decision:
Alt text is very important for screen reader users.
At the same time, features like read aloud do not have similar priority for alt text. In fact some read aloud users like an option to switch off reading of alt text.
So, looking at broad user base of this document, placing alt text after looks better.

clapierre:
Sorry George, looks like you are out voted ;)
I will update the EPUB techniques.

GeorgeKerscher:
I will change the guidelines to have alternative text second.

PR #637, opened 2025-01-30
Minor EPUB technique issues

clapierre:
addressed issues #631, #587, and a missing String Gregorio found this morning.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -363,7 +363,7 @@
                         </li>
                         <li>
                             <span><b>ELSE IF</b> <var>audio_content</var>:</span>
-                            <span><b>THEN</b> display <code id="ways-of-reading-prerecorded-audio-complementary">"Complementary audio and text"</code>.</span>
+                            <span><b>THEN</b> display <code id="ways-of-reading-prerecorded-audio-complementary">"Prerecorded audio clips"</code>.</span>
                         </li>
 
                         <li>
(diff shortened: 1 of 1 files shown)


a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -296,15 +296,14 @@ <h5>Variables setup</h5>
...
                             <span><b>IF</b> <var>all_necessary_content_textual</var>:</span>
                             <span><b>THEN</b> display <code id="ways-of-reading-nonvisual-reading-readable">"Readable in read aloud or dynamic braille"</code>.</span>
                         </li>
+                        <li>
+                            <span><b>IF</b> <var>textual_alternative_images</var>:</span>
+                            <span><b>THEN</b> display <code id="ways-of-reading-nonvisual-reading-alt-text">"Has alternative text"</code>.</span>
+                        </li>
                         <li>
                             <span><b>ELSE IF</b> <var>non_textual_content_images</var> <b>AND</b> <b>NOT</b> <var>textual_alternative_images</var>:</span>
                             <span><b>THEN</b> display <code id="ways-of-reading-nonvisual-reading-not-fully">"Not fully readable in read aloud or dynamic braille"</code>.</span>
```
mattgarrish, line 309: Putting the IF test here turns it into a new set of if/else statements.

The ONIX techniques make this IF a subtest so it's only output if all_necessary_content_textual is true.

If the value should be output no matter the overall readability in braille or using reading aloud, it should be a separate IF after this series of if/else tests. (That would need fixing in both docs.)

Issue #633, opened 2025-01-29
describedMath alone should not be a flag for alternative text


mattgarrish:
As we've discussed before, described math can apply to mathml equations, latex, etc. It's not a statement about the format of the math, so it shouldn't be a flag in the nonvisual reading for saying that alternative text is available for images unless it's paired with a mathOnVisual access mode.

GeorgeKerscher:
Matt writes: 
saying that alternative text is available for images unless it's paired with a mathOnVisual access 
George:

1. Can we specify  this out?
2. Is there any other case where described math works? Possibly if there is MathML and there is alttext attribute filled in?
It seems that the images with alt text is the most common experience we will encounter.

mattgarrish:
Described math isn't only about alt text attributes. You could use aria-details and give an extended description of an equation in a details element. That's why we already took the word "images" out of the string in the rich content description when it's present.

You need to check if accessMode=mathOnVisual is present and pair it with a check if describedMath is set if the descriptive string is going to say images are described.

The alternative is to do what we did in rich content and take "images" out of the descriptive statement so it only says: "Has alternative text descriptions".

PR #639, opened 2025-01-30
Update the metadata testing for nonvisual content display

mattgarrish:
This is my best effort to capture as much as we can about nonvisual reading ability, not accounting for the various edge cases we discussed in the last editor's call.

I've changed the existing variables follows:

- I added a `has_nonvisual_metadata` variable that checks if any accessMode or accessModeSufficient properties are set. If not, then we don't know anything about the content and this triggers the no information available message.
- I modified `all_necessary_content_textual` to add a check that there is only one accessMode property and it has the value textual. This fills in the gap of accessModeSufficient not being required metadata
- I changed the variable `non_textual_content_images` to `non_textual_content` and updated it to check for any accessMode that is not textual so we capture audio and video in addition to images.
- I added a check for transcripts to `textual_alternatives` variable since these could be used for audio or video with images of text.

With those changes, the only major modification I had to make to the instructions is to put the "if" check that has_nonvisual_metadata is not true first. This stops us from commenting on any publications for which the publisher hasn't said anything about the access modes and lets the final else capture the ambiguous cases where they have.

The check on `all_necessary_content_textual` will now better capture all books we know have full text alternatives.

The check on `non_textual_content` without `textual_alternatives` will capture the case that some non-textual content is not going to be accessible. (There are ways this can fail, of course, but it relies on the publisher having only provided partial metadata.)

And finally is the catch-all case where we just don't know. There's non-textual content with some kinds of text alternatives, but for whatever reason the publisher isn't claiming the text is fully readable as text.

But if you spot anything that doesn't seem right, let me know.

Note that I added the new ID `ways-of-reading-nonvisual-no-metadata` for the new "No information about nonvisual reading is available" string. I didn't modify the guidelines to add this.

***

[Preview](https://raw.githack.com/w3c/publ-a11y/fix/epub-nonvis/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/fix/epub-nonvis/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html)

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -254,7 +254,12 @@
                             <p>If true it indicates that there is only a single access mode of textual or <i>accessModeSufficient="textual"</i> (all main content is provided in textual form) is present in the package document, otherwise if false it means that the metadata is not present.</p>
                             <p><i>All non-decorative content supports reading without sight</i> means that all contents of the digital publication necessary to use and understanding, including text, images (via their alternative descriptions), audio and video material (via their transcripts, descriptions, captions or subtitles) are fully accessible via suitable reading devices, for example text-to-speech screen readers or tactile reading devices (‘Braille displays’), and nothing in the digital publication prevents or blocks the use of alternative reading modes. The entire publication can be navigated and ‘read’ using only text rendered via sound or touch, and does not require visual perception.</p>
                         </dd>
-                        <dt><var>non_textual_content</var></dt>
+                    	<dt><var>has_nonvisual_metadata</var></dt>
+                    	<dd>
+                    		<p>If true it indicates that the publisher has stated a least one access mode or sufficient access mode.</p>
+                    		<p>If this metadata is not set, it is not possible to determine whether or not the content will be readable nonvisually.</p>
+                    	</dd>
+                    	<dt><var>non_textual_content</var></dt>
                         <dd>
                             <p>If true it indicates that there is non-textual content such as images, audio, or video in the publication.</p>
                         </dd>
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -335,13 +335,7 @@
                         <dd>
                             <p>If true it indicates that the <i>accessModeSufficient="auditory"</i> (all main content is provided in audio form) is present in the package document, otherwise if false it means that the metadata is not present.</p>
                         </dd>
-                        <dt><var>synchronised_pre_recorded_audio</var></dt>
-                        <dd>
-                            <p>If true it indicates that the <i>accessibilityFeature="sychronizedAudioText"</i> is present in the package document, otherwise, if false, it means that the metadata is not present.</p>
-
-                            <p>This indicates that text-synchronised prerecorded audio narration (natural or synthesized voice) is included for substantially all textual matter, including all alternative descriptions, e.g. via a SMIL media overlay.</p>
-                        </dd>
-
+                        
                         <dt><var>audio_content</var></dt>
                         <dd>
                             <p>If true it indicates that the <i>accessMode="auditory"</i> is present in the package document, otherwise, if false, it means that the metadata is not present.</p>
@@ -349,7 +343,12 @@
                             <p>This indicates that prerecorded audio content is included as part of the work.</p>
                         </dd>
 
+                        <dt><var>synchronised_pre_recorded_audio</var></dt>
+                        <dd>
+                            <p>If true it indicates that the <i>accessibilityFeature="sychronizedAudioText"</i> is present in the package document, otherwise, if false, it means that the metadata is not present.</p>
 
+                            <p>This indicates that text-synchronised prerecorded audio narration (natural or synthesized voice) is included for substantially all textual matter, including all alternative descriptions, e.g. via a SMIL media overlay.</p>
+                        </dd>
                     </dl>
                     <h5>Variables setup</h5>
                     <ol class="condition">
@@ -357,9 +356,9 @@
 
                         <li><b>LET</b> <var>all_content_audio</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessModeSufficient</i>" and normalize-space()="<i>auditory</i>"]</code>.</li>
 
-                        <li><b>LET</b> <var>synchronised_pre_recorded_audio</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessibilityFeature</i>" and normalize-space()="<i>sychronizedAudioText</i>"]</code>.</li>
-
                         <li><b>LET</b> <var>audio_content</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessMode</i>" and normalize-space()="<i>auditory</i>"]</code>.</li>
+
+                        <li><b>LET</b> <var>synchronised_pre_recorded_audio</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessibilityFeature</i>" and normalize-space()="<i>sychronizedAudioText</i>"]</code>.</li>
 
                     </ol>
                     <h5>Instructions</h5>
@@ -374,7 +373,7 @@
                         </li>
                         <li>
                             <span><b>ELSE IF</b> <var>audio_content</var>:</span>
-                            <span><b>THEN</b> display <code id="ways-of-reading-prerecorded-audio-complementary">"Complementary audio and text"</code>.</span>
+                            <span><b>THEN</b> display <code id="ways-of-reading-prerecorded-audio-complementary">"Prerecorded audio clips"</code>.</span>
                         </li>
 
                         <li>
@@ -393,19 +392,6 @@
 
 				<h4>Understanding the variables</h4>
 				<dl>
-					<dt><var>epub_version</var></dt>
-					<dd>
-                        <p>If set, provides the version of the EPUB Accessibility specification the
-                        	publication conforms to.</p>
-					</dd>
-					<dt><var>wcag_version</var></dt>
-					<dd>
-						<p>If set, provides the version of WCAG 2 the publication conforms to.</p>
-					</dd>
-					<dt><var>wcag_version</var></dt>
-					<dd>
-                        <p>If set, provides the WCAG 2 conformance level claimed.</p>
-					</dd>
 					<dt><var>certifier</var></dt>
 					<dd>
 						<p>Returns the description of <i>a11y:certifiedBy</i> (Compliance certification by (name)) if present in the package document, otherwise if false it means that the metadata is not present.</p>
@@ -431,7 +417,15 @@
 						<p>If true it indicates that the <i>conformsTo</i> is present in the package document, otherwise if false it means that the metadata is not present.</p>
 						<p>This means that the publication conforms to some specification.</p>
 					</dd>
-
+					<dt><var>epub_version</var></dt>
+					<dd>
+                        <p>If set, provides the version of the EPUB Accessibility specification the
+                        	publication conforms to.</p>
+					</dd>
+					<dt><var>wcag_version</var></dt>
+					<dd>
+                        <p>If set, provides the WCAG 2 conformance level claimed.</p>
+					</dd>
 				</dl>
 
 				<h4>Variables setup</h4>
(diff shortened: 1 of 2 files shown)


mattgarrish:
Hm, something's gone really wrong with this pull request. I think I may have undone some of my work and reverted to before I opened the new branch. I'll see if I can fix it, but I might have to open a new pull request. For now, ignore all the hundreds of changes it's flagging in the diff after the nonvisual section.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -251,65 +251,75 @@ <h5>Understanding the variables</h5>
...
+                        </li>
+                    	
+                    	<li>
+                    		<b>LET</b> <var>has_nonvisual_metadata</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property=("<i>schema:accessMode</i>", "<i>schema:accessModeSufficient</i>")]</code>.
+                    	</li>
+                    	
+                        <li>
+                            <b>LET</b> <var>non_textual_content</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="<i>schema:accessMode</i>" and not(normalize-space() = "textual")]</code>.
```
clapierre, line 294: non-textual shouldn't it be <>"textual"?

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -251,65 +251,75 @@ <h5>Understanding the variables</h5>
...
+                        	<b>LET</b> <var>all_necessary_content_textual</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[(@property="schema:<i>accessMode</i>" and normalize-space() = "textual" and count(//meta[@property="schema:<i>accessMode</i>"]) = 1) or (@property="schema:<i>accessModeSufficient</i>" and normalize-space()="<i>textual</i>")]</code>.
+                        </li>
+                    	
+                    	<li>
+                    		<b>LET</b> <var>has_nonvisual_metadata</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property=("<i>schema:accessMode</i>", "<i>schema:accessModeSufficient</i>")]</code>.
```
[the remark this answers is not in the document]
mattgarrish: It checks that there is at least one meta tag that has a property of accessMode or accessModeSufficient so we don't attempt to process a value when they haven't stated anything about the modes. The value doesn't matter.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: No, there's a `not(...)` around the value test to capture anything but textual values.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -251,65 +251,75 @@ <h5>Understanding the variables</h5>
...
-                        <dt><var>textual_alternative_images</var></dt>
+                        <dt><var>textual_alternatives</var></dt>
                         <dd>
                             <p>If true it indicates that at least one of the following is present in the package document:</p>
                             <ul>
                                 <li><i>accessibilityFeature="alternativeText"</i> (Short alternative textual descriptions);</li>
                                 <li><i>accessibilityFeature="longDescription"</i> (Full alternative textual descriptions);</li>
                                 <li><i>accessibilityFeature="describedMath"</i> (Visual of math fully described));</li>
-                                <li>otherwise if false it means that this metadata is not present.</li>
+                            	<li><i>accessibilityFeature="transcript"</i> (Transcripts for audio and visual content));</li>
```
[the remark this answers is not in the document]
mattgarrish: Oops, ya, thanks!
```suggestion
                            	<li><i>accessibilityFeature="transcript"</i> (Transcripts for audio and visual content);</li>
```

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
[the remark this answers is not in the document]
clapierre: but if accessMode="visual" is present wont this variable has_nonvisual_metadata be set?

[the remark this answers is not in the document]
mattgarrish: Yes, it's not the variable that checks if the content is all available as text. This is checking if the publisher has set the nonvisual metadata we need to test for completeness.

mattgarrish:
Also, as I wrote in https://github.com/w3c/publ-a11y/issues/633, if we accept this change the descriptive version for alternative text should not include "for images" in it. When the visual content is a video with a transcript, this description would be inaccurate.

gregoriopellegrino:
Can't we remove `ways-of-reading-nonvisual-reading-may-not-fully` at this point?
So instead of having `has_nonvisual_metadata`, we can put `ways-of-reading-nonvisual-no-metadata` in the ELSE.

mattgarrish:
I do hate how we're splitting hairs here between not fully readable and may not be readable.

I'd be more comfortable if we had tests for visual works, audiobooks, and digital braille so that we're not relying on features to try and infer the nature of this content, which will probably end up wrong.

It's easy to add tests that there is only one accessMode and its value is visual, auditory or tactile and push out either:

- Not readable in read aloud or refreshable braille (for purely visual and audiobooks)
- Readable in refreshable braille (for digital braille)

Then I think we can rearrange it so that the end statement is that no information is available.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -286,34 +286,48 @@ <h5>Variables setup</h5>
                         </li>
                     	
                     	<li>
-                    		<b>LET</b> <var>has_nonvisual_metadata</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property=("<i>schema:accessMode</i>", "<i>schema:accessModeSufficient</i>")]</code>.
+                    		<b>LET</b> <var>audio_only_content</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessMode</i>" and normalize-space() = "auditory" and count(//meta[@property="schema:<i>accessMode</i>"]) = 1]</code>.
```
clapierre, line 289: or accessModeSufficient="auditory" right?

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -286,34 +286,48 @@ <h5>Variables setup</h5>
...
                     	</li>
                     	
-                        <li>
+                    	<li>
+                    		<b>LET</b> <var>braille_only_content</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessMode</i>" and normalize-space() = "tactile" and count(//meta[@property="schema:<i>accessMode</i>"]) = 1]</code>.
```
clapierre, line 293: same here for accessModeSufficient="tactile" ideally accessMode= tactile should also be there but just wondering because lets say you have a book all in braille with the tactile images also included. then you would have accessMode = textual and visual, and accessmodesufficient="tactile.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -286,34 +286,48 @@ <h5>Variables setup</h5>
...
                             <b>LET</b> <var>non_textual_content</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="<i>schema:accessMode</i>" and not(normalize-space() = "textual")]</code>.
                         </li>
 
                         <li>
                             <b>LET</b> <var>textual_alternatives</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="<i>schema:accessibilityFeature</i>" and normalize-space()=("<i>longDescription</i>", "<i>alternativeText</i>", "<i>describedMath"</i>, "<i>transcript</i>")]</code>.
                         </li>
+                    	
+                    	<li>
+                    		<b>LET</b> <var>visual_only_content</var> be the result of calling <a href="#check-for-node">check for node</a> on <var>package_document</var>, <code class="xpath">/package/metadata/meta[@property="schema:<i>accessMode</i>" and normalize-space() = "visual" and count(//meta[@property="schema:<i>accessMode</i>"]) = 1]</code>.
```
clapierre: I would personally like to see accessModeSufficient="visual" by itself as that is what we have been telling everyone if this is a picture book that is the main metadata to include.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: No, this is to say that the book is not readable in read aloud or refreshable braille. If we look at sufficient access modes, we could end up catching text publications with media overlays, which we wouldn't want to flag as not readable.

That's not true of visual where you could have only a primary access mode of visual but there's alt text on the images so it has somewhat readable text.

Similarly, if you only check for a textual access mode you'd miss mixed text with visual/auditory publications where the text alternatives are sufficient. That's why we need to check both access mode and access mode sufficient to see if all the content is readable as text.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: Tactile images aren't visual, so you wouldn't flag such a publication as having a visual access mode. That's why the mode is "tactile" and not strictly "braille". No one is expected to view the tactile images visually.

mattgarrish: Even if the producer can't convert the source visual images, wouldn't they toss them out and put in alt text plus descriptions in braille, since they're of no use to braille readers?

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: But that won't tell you it's a visual book with no text alternatives. It just tells you that one path through the content is visual.

If a visual work has alt text or text descriptions, it also has a sufficient access mode of "visual, textual". That means there is some text, although it's not fully equivalent as you still need to see the pictures. This test should probably replace trying to pick out accessibility features. In other words, if textual is only one of a combined sufficient access mode sets, then the publication is not fully readable as text.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
clapierre: Sorry I meant to say a braille book with supplemental visual picture images within it ( believe thats what we were thinking for eBraille)
wouldn't the metadata look like
accessMode = tactile
accessMode = visual
accessModeSufficient = "tactile"

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: I've modified the algorithm to test both. So if there is non-textual content but there is either a textual combined sufficient access mode or the text alternatives, we say not all the content is readable.

Since full text equivalency is captured earlier on, this should capture the rest of the cases where some text is available. It's a bit of an overlapping test, but ensures that if the publisher got at least their sufficient access modes right or listed all their features the publication won't fall through to no information available.

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: I thought we left off needing some way to flag non-tactile images if they weren't going to be banned from braille files. Maybe we should wait and see what happens with that work first?

clapierre: You could be right, thats fine.

mattgarrish:
This is also the point where the visualizer comes in handy. I've made a branch and updated it to match this pull request, and am using that to test what comes out with various permutations of the metadata.

If anyone wants to help out, it's here:  https://raw.githack.com/daisy/a11y-meta-viewer/test/nonvis/index.html

Note that it doesn't care if the metadata or package document are complete, so for testing you can start with:

```
<package xmlns="http://www.idpf.org/2007/opf">
  <metadata>
    <meta property="schema:accessMode">textual</meta>
  </metadata>
</package>
```

and play around from there to see what happens

clapierre (approved):

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: I'm having a hard time with it absent some real examples of what people will do. Is visual actually a sufficient access mode in the case you cite, for example, because the image is not primarily for blind readers but if you can sight read braille you could read the braille and view the images?

I guess a braille book could be transcribed that omits alt text and descriptions for images, but it seems weird, so are alt text and descriptions considered features or are they treated as part of the base tactile access mode?

It makes my head spin and we don't need to solve all the edge cases of braille metadata today!

clapierre: Yeah, I was thinking that you have a braille book that could have tactile images, but a real live image also provided so a teacher could see the image the student was reviewing tactilely. But I don't know if these exist in the wide or something that now with the Monarch multi-line display may become more popular.  I agree we can always revise the guidelines as we get more information on what publishers are putting in their books.

mattgarrish: Ya, or are they some kind of accessibility feature because the user could get an AI-generated description of the image if they don't like or trust what the transcriber provided?

I think what we have is fine for now for catching that the bulk of braille books are good for reading on refreshable braille displays.

mattgarrish:
I've added the new strings to the guidelines. We get one new string for the dynamic braille only output and the not fully string changes to not at all. I adapted the existing string patterns for the descriptive statements, but let me know if anyone wants changes.


***

[Preview](https://raw.githack.com/w3c/publ-a11y/fix/epub-nonvis/a11y-meta-display-guide/2.0/draft/guidelines/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/guidelines/index.html&doc2=https://labs.w3.org/spec-generator/%3Ftype=respec%26url=https://raw.githack.com/w3c/publ-a11y/fix/epub-nonvis/a11y-meta-display-guide/2.0/draft/guidelines/index.html)

gregoriopellegrino:
Sorry, I've been thinking about this for a while. The initial goal of this key accessibility information was to be able to give information to blind people if all the content was usable through the most widely used assistive technologies. We achieve this through `accessModeSufficient = textual`.

Then we had made the discussion that most reflowable ebooks do not have accessibility metadata, but at the same time they are usable for the most part through assistive technologies. On this infer we had doubts.

Now it seems to me that the algorithm has become very complex, going to check edge cases like braille-only publications, which I don't think are present in the trade digital publishing supply chain.

My proposal would be to go back to the simple solution, relying only on the `accessModeSufficient = textual` metadata and indicating that we have no information, when the metadata is not available.

avneeshsingh:
I think that we can avoid braille use cases at this time and make the algorithm simpler. There is demand for considering specialized publication e.g. braille. I think we should do it in the next iteration.

mattgarrish:
If there are only two states, fully readable or no information available, then it at least avoids the weirdness of the old algorithm where audio-, visual-, and braille-only books were assigned misleading statuses because of heuristically trying to figure out readability from some alt text features.

But a lot of books we could say something about based on the access modes and sufficient access modes would be lost to "no information available".

The algorithm as it is is as accurate to the possible stated access modes and sufficient access modes as I can get it. It's more complex than a binary all or nothing, but you can't have simple and accurate and handle the problem of access mode sufficient not being required.

GeorgeKerscher:
I just reviewed the PR. The Guidelines look fine to me. I went through the EPUB techniques, and this also looks good to me.

I think we need to finish this work and get it out to the world as we discussed.

I think we may need to revisit these guidelines when used by libraries serving the blind.  Common DAISY and braille metadata would impact what is presented. However, for the mainstream, I think we are good.

GeorgeKerscher (approved):

gregoriopellegrino (approved):

GeorgeKerscher:
I suggest that we merge this PR. @gregoriopellegrino  can then finalize the excel and JSON.

avneeshsingh (approved):

mattgarrish:
The last commit just removes an unnecessary xpath test that textual isn't the only value in accessmodesufficient when checking if there is some text available. Content with a single sufficient access mode of textual is captured by the check that all text is available, so even if true it will never trigger the wrong output.

gregoriopellegrino:
I am replicating this for ONIX, I realized that the `some_sufficient_text` variable is not explained.

gregoriopellegrino:
I also think this condition should be rewritten:
```
ELSE IF non_textual_content AND (some_sufficient_text OR textual_alternatives)
```
Because it currently returns three only if there are `non_textual_content`, while we could have a pure text book, I would change it to:
```
ELSE IF some_sufficient_text OR textual_alternatives:
```
So it also simplifies and we reduce the variables

mattgarrish:
I'll open another pull request to fix these.

PR #641, opened 2025-02-03
ONIX implementation for Supports nonvisual reading

gregoriopellegrino:
I replicated what I did for EPUB techniques, with some adjustments for the ONIX standard.


mattgarrish (approved):
Looks good to me, although I have almost no onix data to test against. I'm working at merging the two sets of algorithms into a single processing code base, though, and didn't have any trouble combining this with the epub code.

clapierre (approved):

PR #648, opened 2025-02-04
Removed ways-of-reading-nonvisual-reading-no-metadata

GeorgeKerscher:
I found the ID in the definition list and removed the descriptive and compact.

In the note, I removed the (see examples). It did not make sense  to mee to have people  go looking for an example that does not exist.

Not sure how often I will be checking email. Go ahead and merge, unless you see problems.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -716,8 +716,7 @@
 
 				<div class="note">
 					<p>This display field should be rendered even
-						if there is no metadata (See the examples
-						where "No information is available"). </p>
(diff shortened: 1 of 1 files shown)


mattgarrish:
@GeorgeKerscher I opened https://github.com/w3c/publ-a11y/pull/647 to fix the id problem. The techniques output the no information is available statement but the IDs they used were missing part of the string the other nonvisual identifiers have. No need to delete the strings.

I think we can close this without merging if you're okay with that other pull request.

GeorgeKerscher:
Aloha,

If this PR is not needed, feel free to close it. I found those IDs in the guidelines. Of course, if another unmerged PR fixes this, that is fine.

Have we addressed the problem of pointing to examples for no information available?

PR #650, opened 2025-02-05
Adds a version metadata to the canonical json so it will be possible to log and identify vocabulary changes

gautierchomel:
As discussed in editor's call.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
+++ b/a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
@@ -3,7 +3,7 @@
         "author": "W3C Publishing Community Group Accessibility Task Force",
         "language": "en-US",
         "variant": "canonical",
-        "version": "2.0",
+        "version": "2.0.a",
         "audience": "general",
         "description": "Original wording discussed by a large group representing different actors of the English-speaking geographies. It has been improved after proof of concept implementations and panel testers"
     },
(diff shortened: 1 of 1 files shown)


a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json
```diff
@@ -3,6 +3,7 @@
         "author": "W3C Publishing Community Group Accessibility Task Force",
         "language": "en-US",
         "variant": "canonical",
+        "version": "2.0",
```
clapierre, line 6: We want this initial pre-release to be "0.9"

clapierre: Good question though, since the guide is under the 2.0 directory will this initial release be 2.0 or 1.0?

mattgarrish: I would try to keep the version number the same as the guidelines/techniques otherwise it will get confusing.

So, for the 2.0 release, we'd use 2.0.# for updates. Then if we release a 2.1 of the guidelines and techniques it would go to 2.1.#

If that works, we could always call this 2.0.a for an alpha release and then switch to 2.0.0 or 2.0.1 for the first official release.

clapierre: I am fine with 2.0.a for the alpha release but if we need another would it be 2.0.a.2?

GeorgeKerscher:
I agree with keeping techniques and guidelines in sync with numbering of the identification.
2.a.1 seems fine, because we may need to make changes. I like 2.0.0 for the first approved version.

Best
George

mattgarrish (approved):

PR #656, opened 2025-02-14
added back missing Note

clapierre:
This was accidently removed a couple months ago during a major rewrite.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -1098,6 +1098,9 @@
 									data-localization-id="conformance-certifier-credentials"
 									data-localization-mode="compact">The certifier's credential is </span>
 								<span class="placeholder">&lt;certifier credential&gt;</span></p>
+                            <div class="note">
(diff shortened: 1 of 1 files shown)


a11y-meta-display-guide/2.0/draft/guidelines/index.html
```diff
@@ -1098,6 +1098,9 @@ <h5>General conformance</h5>
 									data-localization-id="conformance-certifier-credentials"
 									data-localization-mode="compact">The certifier's credential is </span>
 								<span class="placeholder">&lt;certifier credential&gt;</span></p>
+                            <div class="note">
+								<p>If the certifier has a badge or a credential, then the text is provided. If there is a link to their credential, then this can be provided as a link. If the certifier has a logo or badge, this can be displayed. The display of the logo or badge can be arranged between the certifier and the distributor showing the accessibility metadata.</p>
```
mattgarrish, line 1102: It sounds like there are three scenarios here: text, url, or image. The first two are taken straight from the metadata and the last is a replacement the distributor does based on the text value of the metadata. If so, does it make sense to be a little more explicit, maybe like this:

```suggestion
								<p>If the certifier metadata value is a text-based badge or a credential, then only the text is displayed. If the badge or credential is expressed as a URL, then the URL can be made a link for users to access. The display of image logos or badges is typically arranged between the certifier and the distributor. The distributor would be responsible for identifying the certifier's text credential and replacing it with the image.</p>
```

mattgarrish (approved):

GeorgeKerscher (approved):

gregoriopellegrino (approved):

PR #686, opened 2025-03-31
Add placeholder doc for guidelines implementation report

mattgarrish:
The file to work on will be under the /a11y-meta-display-guide/2.0/implementations/ folder

The URL to it once merged will be https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/implementations/

There's not really anything to review yet as it's just a shell, so you can merge this and start working on it whenever you're ready @GeorgeKerscher

### DIFF
--- a/a11y-meta-display-guide/2.0/implementations/index..html
+++ b/a11y-meta-display-guide/2.0/implementations/index..html
@@ -2,41 +2,58 @@
 <html xmlns="http://www.w3.org/1999/xhtml" lang="en-US" xml:lang="en-US">
 	<head>
 		<meta charset="utf-8" />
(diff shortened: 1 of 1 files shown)


GeorgeKerscher:
I think this is a good start. I find my name there as editor, which is fine. However, There will be a lot of design and CSS considerations that will require a larger group.

One of the first implementations we can show is the viewer Matt developed.

mattgarrish:
I kind of wonder about using respec for this page. It's not a community group report in the usual sense, but using respec requires us to make it one, so have an editors, and abstract, etc. It's not a document that we'd publish in the same way as the guidelines.

Maybe it would be better to export the static html and then trim out all the stuff we don't want?

GeorgeKerscher:
I agree that straight HTML would be better. Less to get in the way.

PR #687, opened 2025-04-01
Fix typos

mickael-menu:
This PR addresses a few typos in the index draft:

- Changed `in it package` to `in its package`
- Changed `distributer` to `distributor`, to be consistent in the document. "Distributer" is an obsolete term.
- Removed an extra parenthesis in `Transcript(s) provided)`
- Removed "No information available" from the [accessibility summary example](https://w3c.github.io/publ-a11y/a11y-meta-display-guide/2.0/draft/guidelines/#sum-examples). It was showing both the state "no information" and an actual content, which was not consistent with the other examples.


clapierre (approved):
Very odd that we had "No information is Available" I wonder if that was a left over from something else. Nice catch.

gautierchomel (approved):

PR #691, opened 2025-04-07
Add entry for the daisy viewer

mattgarrish:
Opening this for input on whether it's sufficient as is or what else we want for these reports.

***

[Preview](https://raw.githack.com/w3c/publ-a11y/impl/daisy-viewer/a11y-meta-display-guide/2.0/implementations/index.html)


clapierre (approved):

mattgarrish:
I had to fix the file name for report (two periods in the extension), so the preview and diff links should now be working.

GeorgeKerscher:
Good start! A few questions:

1. Does the screen shot display a record? If yes, should the output be put in a details element so screen reader users would be able to read the output?
2. Do you want to put in a link to the live viewer online?
3. Do you want to mention that this is an online tool?
4. You say currently accepts , can you expand and say something about a publisher can paste in an ONIX record or an OPF file?
5. I think you should you should say this freely available like all the DAISY tools.

mattgarrish:
@GeorgeKerscher I've made some updates to the section to address your comments. I've added a link to the site and daisy as the owner to a mini info section after the header.

I'm not completely sure what info users are supposed to get from the screenshot, to be honest. I've been treating it as a minor visual aid that people reading this document might look at to get an idea of the formatting. I wasn't thinking they'd try to read the output in the graphic, as it's just a cropped bit of display. I'm assuming they would go to the relevant site to try it out for themselves. Adding a record also means having to maintain it in this site over time. We already have a few (not quite complete) samples for the viewer, so I've just noted those.

GeorgeKerscher:
Excellent improvements! I approve.

GeorgeKerscher (approved):

GeorgeKerscher:
Matt, I think this is good to merge prior to our call on Thursday.

PR #701, opened 2025-04-12
Update all relative paths to use latest link urls

mattgarrish:
This is prep for the final release of the guidelines. It changes all the relative path linking between documents to use the new latest link urls. These urls are not yet active, so they won't work until we finalize the publication date and publish. We can merge this as we get closer to publishing.

(We can't keep the current relative paths as the publishing process requires a flat directory structure and I'm not sure it'd be wise to assume the documents could reference each other using relative paths after publishing even if we tried to tweak them.)

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/guidelines/index.html
+++ b/a11y-meta-display-guide/2.0/draft/guidelines/index.html
@@ -12,9 +12,9 @@
 			group: "publishingcg",
 			specStatus: "CG-DRAFT",
 			noRecTrack: true,
-			latestVersion: "https://www.w3.org/publishing/a11y/UX-Guide-metadata/principles/",
+			latestVersion: "https://www.w3.org/publishing/a11y/metadata-display-guide/guidelines/",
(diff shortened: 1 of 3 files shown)


GeorgeKerscher:
Should we have version 1 of the User Experience now point to the guidelines 2.0?

Should the Accessibility summary guidance  be marked as obsolete and point to the new guidelines?

mattgarrish:
> Should we have version 1 of the User Experience now point to the guidelines 2.0?

No, that can't be done until version 2.0 is published to the w3c servers. We don't want it pointing at the drafts in this repository.

> Should the Accessibility summary guidance be marked as obsolete and point to the new guidelines?

What's the old guidance and what's the new? I'm not sure I know about this change.

mattgarrish:
> No, that can't be done until version 2.0 is published to the w3c servers.

And for this, Ivan will fix the redirects when we tell him the new release is live, same as he'll make these new redirects work. We don't need to change anything in the documents.

clapierre (approved):

PR #706, opened 2025-04-23
Checks printPageNumbers in the EPUB technique

### LINKED ISSUE #703: Legacy `printPageNumbers` should be adressed in the EPUB techniques
As per [Schema.org Accessibility Properties for Discoverability Vocabulary 4.2.1.5 pageBreakMarkers](https://www.w3.org/community/reports/a11y-discov-vocab/CG-FINAL-vocabulary-20241209/#pageBreakMarkers):

> The value `printPageNumbers` is considered a synonym of `pageBreakMarkers`. Although its use is not formally deprecated, authors are strongly encouraged to use the more general `pageBreakMarkers`. 

The Display Techniques for EPUB Accessibility Metadata 2.0 have instructions to process only `pageBreakMarkers`, but should consider both.

gautierchomel:
This proposal should close #703

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -1402,7 +1402,7 @@
 						
 						<dt><var>page_break_markers</var></dt>
 						<dd>
-							<p>If true, it indicates that <i>accessibilityFeature="pageBreakMarkers"</i> is present in the package document.</p>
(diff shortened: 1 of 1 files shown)


clapierre (approved):

gregoriopellegrino (approved):

mattgarrish:
Before merging the current change logs will need to be removed and a new one started for changes since 2.0, plus add this change.

mattgarrish:
Actually, we probably shouldn't work in the 2.0 directory anymore. We should figure out what the next version number will be, copy the drafts, and then re-open this against the new version.

gautierchomel:
To be done in 2.1

PR #708, opened 2025-04-23
Add shortnames to the display guide files

mattgarrish:
We forgot to add these to the respec metadata and when I exported the files to publish respec instead inserted the repository name into the final report URLs.

### DIFF
--- a/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
@@ -8,6 +8,7 @@
 		<script class="remove">
 			// <![CDATA[
 			var respecConfig = {
+				shortName: "epub-techniques",
 				group: "publishingcg",
 				specStatus: "CG-DRAFT",
 				noRecTrack: true,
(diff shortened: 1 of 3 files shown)


clapierre (approved):

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
```diff
@@ -8,6 +8,7 @@
 		<script class="remove">
 			// <![CDATA[
 			var respecConfig = {
+				shortName: "epub-techniques",
```
gautierchomel, line 11: Shouldn't this short name be more explicit?
Like with a `a11y-display-` prefix?

a11y-meta-display-guide/2.0/draft/techniques/onix-metadata/index.html
```diff
@@ -8,6 +8,7 @@
 		<script class="remove">
 			// <![CDATA[
 			var respecConfig = {
+				shortName: "onix-techniques",
```
gautierchomel, line 11: Shouldn't this short name be more explicit?
Like with a `a11y-display-` prefix?

a11y-meta-display-guide/2.0/draft/techniques/epub-metadata/index.html
mattgarrish: This is only for the dated final version that's published on the w3c server. For example:
https://www.w3.org/community/reports/publishingcg/CG-FINAL-onix-techniques-20250422/

It doesn't need to be too specific as this isn't how people will commonly access the guide and techniques.

The current version links that we'll use for referencing are already clear enough:
https://www.w3.org/publishing/a11y/metadata-display-guide/techniques/onix/

PR #710, opened 2025-04-24
Updates for starting version 2.1

### LINKED ISSUE #703: Legacy `printPageNumbers` should be adressed in the EPUB techniques

mattgarrish:
As discussed on the last editor's call, this pull request:

- create a new folder for the 2.1 release and copies and updates all the files from 2.0.
- adds the missing short names to the documents;
- moves the implementations report page to the root guidelines folder so it is not version-specific and adds a redirect on the old page

Some files and folders will be missing after merging because there's no 2.1 equivalent yet (e.g., the published drafts and some of the localization outputs).

@gregoriopellegrino you'll want to check your scripts after we merge this and make sure they all still run.

@gautierchomel I also coped your updates in #706 for printPageNumbers and added a change log entry so this fixes #703


clapierre (approved):

mattgarrish:
I'm just going to go ahead and merge this because if anything's been missed we'll figure it out as we go. Too many new files to pick through.

Issue #721, opened 2025-05-14
Final version has obvious mistakes in conditions


VolodymyrLapan:
Hello,

Your [CG-FINAL-onix-techniques](https://www.w3.org/community/reports/publishingcg/CG-FINAL-onix-techniques-20250422/) document contains following errors:

1. In condition
    IF unknown_motion_simulation_hazard:
      THEN display "Motion simulation hazards not known"[ID: hazards-motion-unknown].
    
    Variable definition for `unknown_motion_simulation_hazard` does not exist.

2. In condition
    IF no_motion_simulation_hazard:
      THEN display "No motion simulation hazards"[ID: hazards-motion-none].

    Variable definition for `no_motion_simulation_hazard` does not exist.

3. In condition
    IF NOT (math_formula_as_mathml OR math_formula_as_latex OR (contains_math_formula AND short_textual_alternative_images) OR 
    chemical_formula_as_mathml OR charts_diagrams_as_non_graphical_data OR full_alternative_textual_descriptions OR closed_captions OR 
    open_captions OR transcript):
    THEN either omit this display field if metadata is missing or display "No information is available"[ID: rich-content-unknown].

Variable definition for `short_textual_alternative_images` and `contains_math_formula` does not exist.

And please, add some explanation how things like `normalize-space()` or `(@lang|ancestor::*/@lang)[last()` should work, most parsers does not support it. We need to know how to implement these functions.

Thanks

mattgarrish:
> 1\. Variable definition for `unknown_motion_simulation_hazard` does not exist.
> 2\. Variable definition for no_motion_simulation_hazard does not exist.

These look like simple typos. Maybe from copying from the epub techniques. The variables are unknown_motion_hazard  and no_motion_hazard everywhere else.

> 3\. Variable definition for `short_textual_alternative_images` and `contains_math_formula` does not exist.

This one I'm not sure about. @gregoriopellegrino should this part of the test be removed or are there variable definitions missing?

> normalize-space() or (@lang|ancestor::*/@lang)[last()

These are xpath expressions. normalize-space trims and compacts whitespace within the given value (the text content of the selected node).

The node selector gets the language declaration that is currently in scope for the element. That will either come from a lang attribute on the selected node, if there is one, or the nearest ancestor element that has a lang attribute. The selector gets all of them and the last() call returns the one nearest to the element.

PR #727, opened 2025-06-02
Fixing issues in #721

gregoriopellegrino:


mattgarrish (approved):
This looks fine.

Explaining the xpaths we'd need to discuss as it would take a lot of work to do. We did something similar with the publication manifest processing using details elements.

mattgarrish:
But it's also time-consuming to do and I don't know that we have the time right now to do it with everything else we want to work on.

PR #730, opened 2025-06-12
Rename guide title

mattgarrish:
Switches "EPUB Accessibility Metadata Guide" to "EPUB Accessibility &#8212; Schema.org Metadata Guide"

(An alternative would be "Package Document Metadata Guide" but that would only be needed if we were covering conformance metadata, which I don't think this guide was expected to get into?)


mattgarrish:
> Package Data Metadata Guide

Data and Metadata suggest the same thing. And now I wonder if we should use the "EPUB Accessibility" label in the CG. Maybe it would be safer to make it something like:

Guide to Expressing Accessibility Metadata in the EPUB Package Document

We could probably leave the short name and directory alone in this case, too. Does that work?

clapierre:
Yeah, I kind of like "Expressing Accessibility Metadata in the EPUB Package Document"

GeorgeKerscher:
I too like "Expressing Accessibility Metadata in the EPUB Package Document"

Issue #688, opened 2025-04-02
Localized placeholders


mickael-menu:
I believe there is an issue regarding how placeholders are handled in display statements. The technique documents contain basic concatenation instructions to handle placeholders:

<img width="600" alt="Image" src="https://github.com/user-attachments/assets/b4708d44-d7e8-43cf-b401-eed59ca5d541" />

---

This works fine for English, but we can't guarantee the placeholder will always be located at the end of the sentence. Here's an example translating this sentence in Japanese:

<img width="400" alt="Image" src="https://github.com/user-attachments/assets/30c64646-03f1-48f5-8f78-e4ddbcca7ca4" />

I suggest instead using actual placeholders in the strings. Positional placeholders (e.g. `$1`) are fine, but named placeholders are easier to understand by translators and less error-prone. 

I don't know what type of placeholders does GitLocalize support, but it could be something easy-to-read like that:

`The publication was certified on {date}`

Related issue: https://github.com/w3c/publ-a11y/issues/460

HadrienGardeur:
I fully agree with @mickael-menu, to make these strings usable we would need interpolated strings as suggested above.

We also need to make sure that some of those strings (date for instance) are localized by the implementer. Displaying an ISO8601 date to the user, would make very little sense.

gautierchomel:
Editors agreed it would be the right path but stated that it was a significant change to make and decided to defer to the next version.

avneeshsingh:
The topic was discussed in May 22, 2025 accessibility task force meeting.
Kind attention implementers: @rickj , @gregoriopellegrino , @gautierchomel 

Topic: Next steps for localized placeholders:

AvneeshSingh: https://github.com/w3c/publ-a11y/issues/688

CharlesL: when putting a date in a sentence, depending on the language you need to move the place of the word.

gpellegrino: instead of concatenating, it’s better to have a placeholder in the string, usually between brackets so it can be freely moved by the person localizing.

Hadrien: this came out when implementing in the readium mobile toolkits. It's possible to have more than one variable (day, month, year per example). Interpolating strings necessitate to have each piece in brackets.

AvneeshSingh: that mean a change in the JSON file and a warning for implementers.

Hadrien: we should recommend using interpolated strings and localized dates. As a user I can set up a language and a date format, that may be different. Implementers should not show raw data for dates, but format them to the user preference.

CharlesL: that also affects WCAG versions and levels.

mgarrish: for WCAG we can rely on official translations as they exist.

[...]

PR #732, opened 2025-07-08
First draft for placeholder

gregoriopellegrino:
As discussed in issue #688

### DIFF
--- a/a11y-meta-display-guide/2.1/draft/techniques/onix-metadata/index.html
+++ b/a11y-meta-display-guide/2.1/draft/techniques/onix-metadata/index.html
@@ -1073,7 +1073,7 @@
 					</li>
 					<li>
 						<span><b>IF</b> <var>publisher_contact_for_accessibility</var> is <b>NOT</b> empty:</span>
-						<span><b>THEN</b> display <code id="accessibility-summary-publisher-contact">"For more information about the accessibility of this product, please contact the publisher: {publisher_contact_for_accessibility}"</code> replacing <code>"{publisher_contact_for_accessibility}"</code> with <var>publisher_contact_for_accessibility</var> as link <b>IF</b> <var>publisher_contact_for_accessibility</var> is an email address.</span>
+						<span><b>THEN</b> display <code id="accessibility-summary-publisher-contact">"For more information about the accessibility of this product, please contact the publisher: {publisher_contact_for_accessibility}"</code> replacing <code>"{publisher_contact_for_accessibility}"</code> with <var>publisher_contact_for_accessibility</var> as a mailto link.</span>
 					</li>
 				</ol>
 			</section>


mattgarrish:
Just on a skim this looks fine, but it's a headache to implement to test. If I switch over to placeholders, either I have to duplicate the code to continue to account for the 2.0 files or I'm back to only being able to display English.

We should try and hurry this revision along after adding, perhaps not taking up any additional issues, so we can deprecate 2.0 and get the translations updated.

mattgarrish:
One issue I noticed that's not directly covered here is the certifier's report. The algorithm says to replace the entire string with a hyperlink, but the examples in the guidelines show only the words "certifier's report" hyperlinked. The guidelines approach makes more sense, but is there a way to accommodate that with a placeholder since it's not possible to know what words to hyperlink in any other language.
.
It's similar to these placeholders, especially the email link, except instead of replacing with a variable value we want the marked text hyperlinked. Maybe something other than brackets would work to identify the linkable text, like percent signs?

a11y-meta-display-guide/2.1/draft/techniques/onix-metadata/index.html
```diff
@@ -1058,11 +1064,7 @@ <h4 id="accessibility-summary-instructions">Instructions</h4>
 					</li>
 					<li>
 						<span><b>IF</b> <var>publisher_contact_for_accessibility</var> is <b>NOT</b> empty:</span>
-						<span><b>THEN</b></span>
-						<ol class="display">
-							<li>display <code id="accessibility-summary-publisher-contact">"For more information about the accessibility of this product, please contact the publisher: "</code></li>
-							<li>display <var>publisher_contact_for_accessibility</var> as link <b>IF</b> <var>publisher_contact_for_accessibility</var> is an email address.</li>
-						</ol>
+						<span><b>THEN</b> display <code id="accessibility-summary-publisher-contact">"For more information about the accessibility of this product, please contact the publisher: {publisher_contact_for_accessibility}"</code> replacing <code>"{publisher_contact_for_accessibility}"</code> with <var>publisher_contact_for_accessibility</var> as link <b>IF</b> <var>publisher_contact_for_accessibility</var> is an email address.</span>
```
mattgarrish, line 1067: It should be obvious, but it might be better to say to replace with a "mailto link" here.
```suggestion
						<span><b>THEN</b> display <code id="accessibility-summary-publisher-contact">"For more information about the accessibility of this product, please contact the publisher: {publisher_contact_for_accessibility}"</code> replacing <code>"{publisher_contact_for_accessibility}"</code> with <var>publisher_contact_for_accessibility</var> as a mailto link <b>IF</b> <var>publisher_contact_for_accessibility</var> is an email address.</span>
```

mattgarrish: Also, can the last "IF ... is an email address." be dropped? Doesn't onix require an email address?

mattgarrish:
I've finished implementing this now. The only other slightly tangential comment I have is that maybe we should say something about localizing the date, not just to replace the placeholder with whatever value you get from the metadata. No matter the language it may need some processing, if not to turn the date into the local language then to process out an attached time.

It could just go in the variable setup description, but it's a unique case of localizing the publisher data.

gautierchomel (approved):

a11y-meta-display-guide/2.1/draft/techniques/onix-metadata/index.html
```diff
@@ -586,47 +578,61 @@ <h4 id="conformance-instructions">Instructions</h4>
 						</div>
 						
 						<span><b>IF</b> <var>epub_accessibility_10</var> <b>OR</b> <var>epub_accessibility_11</var> <b>OR</b> <var>wcag_20</var> <b>OR</b> <var>wcag_21</var> <b>OR</b> <var>wcag_22</var> <b>OR</b> <var>level_aaa</var> <b>OR</b> <var>level_aa</var> <b>OR</b> <var>level_a</var>:</span>
```
mattgarrish, line 580: @clapierre made me aware that if you don't have complete metadata this section will generate some weird statements like "This publication claims to conform to EPUB Accessibility 1.1 Level AA" or even just "This publication claims to conform to Level AA" after processing the placeholders.

To avoid this, the test here might need to be modified to be more thorough in what is expected. To check that we get complete statements it could be:

> IF (epub_accessibility_10 OR epub_accessibility_11) AND (wcag_20 OR wcag_21 OR wcag_22) AND (level_aaa OR levela_aa OR level_a):

If we really need to support wcag conformance statements without accompany epub accessibility conformance, then it gets harder as you'd have to test that the epub statements aren't also accompanied by partial wcag statements.

gautierchomel:
Waiting for @mickael-menu & @HadrienGardeur feedback as they originated the issue.

PR #736, opened 2025-07-25
consistently spell ruby with a lower case "r"

### LINKED ISSUE #735: Ruby annotation case inconsistency
In the case of "Full ruby annotations", the word "ruby" starts in lower case:

https://github.com/w3c/publ-a11y/blob/545b874fd311900852671f58f712cdbd8e7e4bad/a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json#L270-L273

and in the case of "Some Ruby annotations", it starts in upper case:

https://github.com/w3c/publ-a11y/blob/545b874fd311900852671f58f712cdbd8e7e4bad/a11y-meta-display-guide/2.0/draft/localizations/en-US/display_guide_vocabulary_w3c_en-US.json#L290-L293

which is inconsistent.

gautierchomel:
closes #735

### DIFF
--- a/a11y-meta-display-guide/2.1/draft/techniques/epub-metadata/index.html
+++ b/a11y-meta-display-guide/2.1/draft/techniques/epub-metadata/index.html
@@ -1506,7 +1506,7 @@
 						</li>
 						<li>
 							<span><b>IF</b> <var>ruby_annotations</var>:</span>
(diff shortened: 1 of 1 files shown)


mattgarrish (changes requested):
This is fixing the old draft 2.0 localization file that's no longer in use.

The change should be made to these strings in the 2.1 techniques. The 2.1 localization file would then be compiled from those.

clapierre (approved):

PR #747, opened 2025-10-17
fixes #746

### LINKED ISSUE #746: Structural navigation doesn't match schema.org definition
The explanation in the epub metadata guide makes it sound more like SC 1.3.1 info and relationships:

> The use of headings, lists, tables, asides, etc. in the resource fully and accurately reflects the document hierarchy, allowing navigation by assistive technologies.

By contrast the official definition only says:

> The use of headings in the resource fully and accurately reflects the document hierarchy, allowing navigation by assistive technologies.

We can argue that "structuralNavigation" as a term for heading navigation is a bit misleading, and I'd agree, but heading structure is more verifiable than a definition that relies on everything being structured properly. At the very least, we would need to redefine the schema.org definition if we want to use the current explanation.

clapierre:
focus primarily on headings and order and added an example

### DIFF
--- a/epub-a11y-meta-guide/1.0/draft/index.html
+++ b/epub-a11y-meta-guide/1.0/draft/index.html
@@ -2542,11 +2542,7 @@ body {
 
 						<p><b>Example Package Document (<code>package.opf</code>):</b></p>
 
-						<pre>&lt;!-- Declaring 'none' means no other accessibility features are claimed. --&gt;
-    &lt;meta property="schema:accessibilityFeature"&gt;none&lt;/meta&gt;
-
-    &lt;meta property="schema:accessibilitySummary"&gt;This publication does not include any specialized accessibility features.&lt;/meta&gt;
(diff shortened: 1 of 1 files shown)


epub-a11y-meta-guide/1.0/draft/index.html
```diff
@@ -961,50 +961,44 @@ <h5>Reading Order</h5>
...
-						</ul>
-
-						<p>The metadata to include in the package document when proper structural elements are used to
-							facility accessible navigation would be:</p>
-
-						<pre>&lt;meta property="schema:accessibilityFeature"&gt;structuralNavigation&lt;/meta&gt;</pre>
-					</section>
-
+                    <section id="metadata-a11y-features-structuralNavigation">
+                        <h5>structuralNavigation</h5>
```
mattgarrish: This switches from an English title to the term name.

We could use a different name, too, to avoid the ambiguity of "structural navigation" as it's more like "heading structure in the markup navigation"

PR #796, opened 2026-03-04
Add additional context for audio via media overlays

### LINKED ISSUE #794: The synchronizedAudioText feature
### Section

https://w3c.github.io/publ-a11y/package-metadata-authoring-guide/#identify-sync-audio-text

### Describe the problem

The guide says:

> To determine if an EPUB 3 publication should declare the `synchronizedAudioText` feature, check if the `media-overlay` attribute has been set on any items in the package document.

A few questions arise:

1\. Does this feature require that audio narration is synchronized to actual (unicode) text?

Apparently not, as indicated here: https://github.com/w3c/publ-a11y/issues/789#issuecomment-3915186313. However, we have seen publishers omit synchronizedAudioText for books where audio is synchronized to fixed layout page images (including images of text), presumably because "text" is interpreted as computer-encoded text.

If that's a misinterpretation, it may be worth stating explicitly.

2\. Does this feature require that most or all of the narrated content is also included as text (either unicode or images of text)?

Apparently it does, but the criterion mentioned (presence of media-overlay attribute) does not guarantee this. The guide itself provides a counterexample (see 2.2.2.3.1 Synchronized text and audio): synchronization limited to major headings is not considered sufficient to qualify for the feature, even though media overlays are present.

This may be a symptom of a structural issue: Synchronized audio and text is discussed in more detail in 2.2.2.3.1 Synchronized text and audio, in relation to the auditory access mode, and then again under 3.2 Aural rendering. The section about the synchronizedAudioText accessibility feature, on the other hand, is very bare-bones and does not really explain what the feature is about. Repeating some essential information there and/or adding cross-references might help.

3\. Does this feature require that most or all of the text is narrated?

Apparently not. The corresponding ONIX value does, however, require synchronized audio for (substantially) all text, and the current display guide and EPUB display techniques rely on that interpretation.

This point is mostly about internal consistency between the different documents and metadata formats and the potential for misinterpretation it creates. I'm not suggesting that Schema vs. ONIX should be discussed here.

### Describe the fix or new feature you propose

_No response_

mattgarrish:
Clarifies the issue about whether text means unicode character data and adds additional context and linking from the sections where audio is discussed to the subsection that details media overlays relative to access modes and sufficient access modes.

The pull request also clarifies that not all tts tech have to be used when setting ttsMarkup. I noticed that wasn't clear while reading the aural rendering features section.

Fixes #794 

***

[Preview](https://raw.githack.com/w3c/publ-a11y/editorial/mo-review/package-metadata-authoring-guide/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/package-metadata-authoring-guide/index.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fpubl-a11y%2Feditorial%2Fmo-review%2Fpackage-metadata-authoring-guide%2Findex.html)

(Edit: the diff should now be working again.)

### DIFF
--- a/package-metadata-authoring-guide/index.html
+++ b/package-metadata-authoring-guide/index.html
@@ -1246,7 +1246,8 @@
 								synchronized text-audio playback.</p>
 
 							<p>For example, a publication with media overlays that has text and images (with text
-								alternatives) would only declare the following sufficient access modes:</p>
+								alternatives sufficient to understand their content) would only declare the following
+								sufficient access modes:</p>
 
 							<pre>&lt;meta property="schema:accessModeSufficient">textual&lt;/meta>
 &lt;meta property="schema:accessModeSufficient">auditory&lt;/meta>
@@ -1254,6 +1255,10 @@
 
 							<p>It would <strong>not</strong> include declarations for "<code>textual,auditory</code>" or
 									"<code>textual,visual,auditory</code>".</p>
+
+							<p>(Note that the final "<code>textual,visual</code>" set captures the two access modes
+								needed to read the default content &#8212; <code>textual</code> for the text content and
+									<code>visual</code> for the images.)</p>
 						</div>
 
 						<div class="note">


package-metadata-authoring-guide/index.html
```diff
@@ -1231,7 +1236,25 @@ <h5>Auditory</h5>
...
+							<p>When media overlays are present, the "<code>auditory</code>" value should not be added to
+								all the possible sufficient access modes just because it is possible to turn on
+								synchronized text-audio playback.</p>
+
+							<p>For example, a publication with media overlays that has text and images (with text
+								alternatives) would only declare the following sufficient access modes:</p>
+
+							<pre>&lt;meta property="schema:accessModeSufficient">textual&lt;/meta>
+&lt;meta property="schema:accessModeSufficient">auditory&lt;/meta>
+&lt;meta property="schema:accessModeSufficient">textual,visual&lt;/meta></pre>
```
clapierre: May want to add why this combo exists.

PR #808, opened 2026-04-21
Rewrite of the features appendix

### LINKED ISSUE #807: The alternativeText feature
### Section

https://w3c.github.io/publ-a11y/package-metadata-authoring-guide/#image-description, https://w3c.github.io/publ-a11y/package-metadata-authoring-guide/#identify-alt

### Describe the problem

A common case where guidance would be helpful is the following: there are no images in the actual work (or all images are decorative), but alt text has been supplied for the cover and possibly for corporate logos and similar content. Should the publication declare alternativeText?

The cover image and publisher-added extraneous visual content are excluded from consideration when determining accessMode visual. Should they also be excluded when determining whether alternativeText should be set?

I would also like to point out that the description of this feature under C.1 Alternative text seems slightly misleading:

> It is generally easy to determine if XHTML content documents have alternative text, as it only requires checking that an HTML alt attribute is present on each non-decorative img tag.

Of course, the presence of an alt attribute is not enough: the attribute must contain meaningful alternative text. In addition, whether an image is decorative can't be inferred from the img tag itself. This requires evaluating the image in its context.

### Describe the fix or new feature you propose

_No response_

mattgarrish:
I've gone over the appendix again to try and better clarify that it's not a guide to setting the feature values.

I've relabeled the appendix "Introduction to accessibility features". "Identifying accessibility features" sounds too much like you can use it as a guide to when to set the metadata. I've also gone over all the features and made them more explanatory in nature.

Hopefully, this better separates the body - which is all about how to determine when the metadata applies - from the appendix - which was only meant to show examples of what the features correspond to in the content.

Fixes #807 

***

[Preview](https://raw.githack.com/w3c/publ-a11y/editorial/clarify-feature-find/package-metadata-authoring-guide/index.html) | [Diff](https://services.w3.org/htmldiff?doc1=https://www.w3.org/publications/spec-generator/%3Ftype=respec%26url=https://w3c.github.io/publ-a11y/package-metadata-authoring-guide/index.html&doc2=https:%2F%2Fwww.w3.org%2Fpublications%2Fspec-generator%2F%3Ftype=respec%26url=https:%2F%2Fraw.githubusercontent.com%2Fw3c%2Fpubl-a11y%2Feditorial%2Fclarify-feature-find%2Fpackage-metadata-authoring-guide%2Findex.html)

### DIFF
--- a/package-metadata-authoring-guide/index.html
+++ b/package-metadata-authoring-guide/index.html
@@ -13,7 +13,7 @@
 				noRecTrack: true,
                 edDraftURI: "https://w3c.github.io/publ-a11y/package-metadata-authoring-guide/",
 				latestVersion: "https://www.w3.org/2026/publ-a11y/latest/package-metadata-authoring-guide/",
-				// thisVersion: "https://www.w3.org/community/reports/a11y-discov-vocab/CG-FINAL-vocabulary-20260326/",
+				// thisVersion: "https://www.w3.org/community/reports/publ-a11y/CG-FINAL-package-metadata-authoring-guide-20260101/",
                 copyrightStart: "2025",
                 editors:[ {
 					name: "Matt Garrish",
@@ -4666,12 +4666,12 @@
 				</dl>
 			</section>
 		</section>
-		<section id="features-identifying" class="appendix">
-			<h2>Identifying accessibility features</h2>
+		<section id="features-intro" class="appendix">
+			<h2>Introduction to accessibility features</h2>
 
-			<p>This appendix provides guidance on how to determine if an <a data-cite="epub-3#dfn-epub-publication">EPUB
-					publication</a> [[epub-3]] contains accessibility features for anyone unfamiliar with their markup
-				or styling.</p>
+			<p>This appendix provides introductory information on what accessibility features are and how to determine
+				if an <a data-cite="epub-3#dfn-epub-publication">EPUB publication</a> [[epub-3]] contains them. It is
+				only for intended for individuals unfamiliar with their markup or styling.</p>
 
 			<div class="caution">
 				<p>Finding features in the markup does not always mean the corresponding metadata applies to a
@@ -4680,13 +4680,13 @@
 						href="#accessibilityFeature">usage requirements</a> for a property before declaring it.</p>
 			</div>
 
-			<section id="identify-alt">
+			<section id="intro-alt">
 				<h3>Alternative text</h3>
 
-				<p>Alternative text provides a brief description of an image.</p>
+				<p><a href="#alternativeText">Alternative text</a> provides a brief description of an image.</p>
 
-				<p>For <a data-cite="epub-3#dfn-xhtml-content-document">XHTML content documents</a> [[epub-3]]
-					this text is found in [[html]] [^img/alt^] attributes on non-decorative [^img^] tag.</p>
+				<p>For <a data-cite="epub-3#dfn-xhtml-content-document">XHTML content documents</a> [[epub-3]] this text
+					is found in [[html]] [^img/alt^] attributes on non-decorative [^img^] tag.</p>
 
 				<aside class="example" title="HTML img tag with alternative text">
 					<pre>&lt;img src="images/hawk.jpg" 
@@ -4705,15 +4705,13 @@
    …
 &lt;/svg></pre>
 				</aside>
-
-				<div class="note">
-					<p>Refer to the <a href="#alternativeText"><code>alternativeText</code> feature</a> for the
-						specifics of when it applies.</p>
-				</div>
 			</section>
 
-			<section id="identify-aria">
+			<section id="intro-aria">
 				<h3>ARIA structural roles</h3>
+
+				<p><a href="#ARIA">ARIA structural roles</a> provide additional context for assistive technologies about
+					the type of content the user is reading.</p>
 
 				<p>ARIA roles are expressed in <a data-cite="epub-3#dfn-xhtml-content-document">XHTML content
 						documents</a> [[epub-3]] using the [[html]] [^/role^] attribute:</p>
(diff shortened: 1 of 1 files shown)


package-metadata-authoring-guide/index.html
```diff
@@ -4896,78 +4941,97 @@ <h3>LaTeX formatting</h3>
...
-					as <code>m:math</code> or <code>mml:math</code>. This kind of tagging is more common in legacy
-					content as <a data-cite="epub-3#dfn-epub-reading-system">reading systems</a> [[epub-3]] used to rely
-					on the MathJAX application to render equations. Modern reading systems, like browsers, expect
-					unprefixed elements as part of their support for HTML.</p>
+				<div class="note">
+					<p>Because EPUB is XML-based, the <code>math</code> tag could also have a prefix (e.g.,
+							<code>m:math</code> or <code>mml:math</code>). This kind of tagging is more common in legacy
+						content as <a data-cite="epub-3#dfn-epub-reading-system">reading systems</a> [[epub-3]] used to
+						rely on the MathJAX application to render equations. Modern reading systems, like browsers,
+						expect unprefixed elements as part of their support for HTML.</p>
```
clapierre: I think we need to make sure to say that we recommend against MathML name spacing even though it may appear in legacy content, 

IE> KB
 Should I prefix the MathML tags?
No, it is not recommended to use prefixes with math tags (e.g., m:math).

Not all combinations of assistive technologies and user agents are XML aware, so even if they support MathML markup they may not support prefixed markup.

The use of unprefixed tags is also recommended as it ensures compatibility between both HTML and XHTML syntaxes (XML prefixes are not supported in HTML).

package-metadata-authoring-guide/index.html
```diff
@@ -4896,78 +4941,97 @@ <h3>LaTeX formatting</h3>
...
 				<h3>Page break markers</h3>
 
-				<p>To check if the <a href="#pageBreakMarkers"><code>pageBreakMarkers</code></a> feature applies, search
-					for the <a data-cite="dpub-aria#doc-pagebreak"><code>doc-pagebreak</code> role</a>
+				<p><a href="#pageBreakMarkers">Page break markers</a> provide users the ability to locate their position
+					relative to a pre-paginted equivalent edition, such as a hard or soft cover version of a book. They
+					also provide link destinations for a <a href="intro-page-nav">page list</a>.</p>
+
+				<p>Page break markers can either be visible or invisible elements in the content. They are identifiable
+					in the markup by having a <a data-cite="dpub-aria#doc-pagebreak"><code>doc-pagebreak</code> role</a>
```
clapierre, line 5036: To help make the examples clear maybe 2 headers above each code snipit, 
```
<h?>Hidden Page Number</h?>
...
<h?>Visible Page Number</h?>
```
The two just looks merged as one

package-metadata-authoring-guide/index.html
```diff
@@ -4979,16 +5043,20 @@ <h3>Page break markers</h3>
...
 
-				<p>For an EPUB 3 publication, check the <a data-cite="epub-3#dfn-epub-navigation-document">EPUB
-						navigation document</a> [[epub-3]] for a [[html]] [^nav^] element with the <a
-						data-cite="epub-3#sec-epub-type-attribute"><code>epub:type</code> attribute</a> [[epub-3]] value
-						"<a data-cite="epub-ssv#page-list"><code>page-list</code></a>" [[epub-ssv]].</p>
+				<p>In an EPUB 3 publication, the page list is found in the <a
+						data-cite="epub-3#dfn-epub-navigation-document">EPUB navigation document</a> [[epub-3]]. It will
+					be contained an [[html]] [^nav^] element with its <a data-cite="epub-3#sec-epub-type-attribute"
+							><code>epub:type</code> attribute</a> [[epub-3]] value "<a data-cite="epub-ssv#page-list"
+							><code>page-list</code></a>" [[epub-ssv]].</p>
```
clapierre, line 5062: I would make the example in the page list hidden OL 

I have seen too many instances where the page list is visible, I know the nav doc shouldn't be part of the spine but you would be surprised and what I have seen :).
`<nav epub:type="page-list" hidden="hidden">`

package-metadata-authoring-guide/index.html
```diff
@@ -5030,46 +5098,57 @@ <h3>Page list</h3>
...
 
-				<p>Checking for structural navigation requires inspecting each heading to ensure that it is represented
-					using one of the [[html]] <a data-cite="html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements"
-							><code>h1</code> through <code>h6</code> elements</a>.</p>
+				<p><a href="#structuralNavigation">Structural navigation</a> is the ability for users of assistive
+					technologies to skip from one heading to the next without having to go through the table of contents
+					every time. It is enabled by having all the headings marked up at their correct level.</p>
+
+				<p>HTML provides the <a data-cite="html#the-h1,-h2,-h3,-h4,-h5,-and-h6-elements"><code>h1</code> through
+							<code>h6</code> elements</a> for marking up headings.</p>
```
clapierre, line 5154: since the number one rule of ARIA is not to use ARIA, I think our 2nd example here should at least be for a heading level not covered by H1-6 so maybe a level="7" ?

package-metadata-authoring-guide/index.html
mattgarrish: This isn't a content authoring guide, though. It's just exhibiting the content for metadata practitioners, and in this case noting that it may appear in different ways for historical reasons. I've tweaked the note to try and make it clearer that prefixes are only likely in legacy content, but that's as far as I'd go.

package-metadata-authoring-guide/index.html
```diff
@@ -5341,6 +5569,9 @@ <h3>Word segmentation</h3>
   margin-right: 0.5em;
 }</pre>
 				</aside>
+
+				<p>A more reliable way to check for word segmentation is to open a publication in a reading system and
+					see if there is spacing between words.</p>
```
clapierre, line 5577: In the last section Terms defined by reference,
Why is "EPUB navigation document" repeated 3 times?

...
defines the following:
alternate-script property
author names
EPUB content document
EPUB navigation document
EPUB navigation document
EPUB navigation document 
EPUB publication

clapierre (approved):

package-metadata-authoring-guide/index.html
mattgarrish: HTML doesn't have seven levels of heading, so there's no guarantee what happens once you create a level beyond 6. ARIA shouldn't be used to create new semantics.

I don't know if it's changed, but Jaws used to treat anything above 6 as an h2.

Anyway, it's not worth getting into these kinds of issues in a metadata guide. Let's leave it as something we know is valid.

clapierre:
Nicely done Matt.

package-metadata-authoring-guide/index.html
clapierre: Yeah, I guess so, just bugs me that we are showing an example where I would flag this saying it should be done using real HTML established semantics instead of trying to invent your own, but I hear you.

clapierre (approved):

clapierre (approved):

package-metadata-authoring-guide/index.html
mattgarrish: I've changed the example so it doesn't look like an alternative representation of the h2 markup that came before it. The aria attributes wouldn't exist if there weren't cases where they're needed, so we don't need to question the appropriateness of its use in this document. I'm just concerned that if we use something like a level 7 it'll invite issues from people who don't think it's valid or will want to point out that it doesn't work everywhere, which is all off-topic for this guide.

package-metadata-authoring-guide/index.html
mattgarrish: Ya, that's interesting. Never seen that behaviour before. The three references all pointed to different places - one to the term definition, one to the nav doc section, and the last one was actually a broken link. Respec seems to treat them as different terms because of the different links, despite all having the same name. I've fixed it so they all refer to the definition of the nav doc now.

